PR #15, opened 2025-12-09
minor: draft loadBalancer

andreasfoo:
1. Unified API Style Adapter

The adapter now supports Anthropic and OpenAI endpoints seamlessly. You can call either provider without caring about the underlying model or service differences.

2. Multi-Service Load Balancing

- A new load-balancing mechanism has been implemented with four tactics:

- Round-robin (default): Switches services when the request-time limit is reached.

- Random: Chooses a service randomly for each request.

- Token-based: Switches services when total consumed tokens reach a configured limit.

- Hybrid: Switches services when either the token limit or request-time limit is reached.

### DIFF
--- a/internal/config/server_config.go
+++ b/internal/config/server_config.go
@@ -76,6 +76,9 @@ func NewConfigWithDir(configDir string) (*Config, error) {
 					ResponseModel: "",
 					Services:      []Service{}, // Empty services initially
 					Active:        true,
+					// Legacy fields for backward compatibility
+					Provider:     "",
+					DefaultModel: "",
 				},
 			}
 			cfg.DefaultRequestID = 0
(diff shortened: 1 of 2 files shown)


internal/config/server_config.go
```diff
@@ -74,8 +74,7 @@ func NewConfigWithDir(configDir string) (*Config, error) {
 				{
 					RequestModel:  "tingly",
 					ResponseModel: "",
-					Provider:      "",
-					DefaultModel:  "",
+					Services:      []Service{}, // Empty services initially
 					Active:        true,
```
FFengIll, line 78: keep these fields for compatibility, and remove in future.

internal/config/server_config.go
```diff
@@ -254,9 +256,9 @@ func (c *Config) GetRequestConfigByRequestModel(modelName string) *Rule {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
-	for _, rc := range c.Rules {
-		if rc.RequestModel == modelName {
-			return &rc
+	for i := range c.Rules {
+		if c.Rules[i].RequestModel == modelName {
+			return &c.Rules[i]
 		}
```
FFengIll, line 265: check the for loop and naming.

PR #24, opened 2025-12-11
feat: migrate to use swagger (openapi v2) for code gen and testing.

FFengIll:
# Major: support swagger.json and codedgen
This PR provide a swagger wrapper (golang module) to help define service endpoints (via gin) and generate swagger.json automatically.
Then we can use it to generate our server and client code.

# Major: define all api mdoels
For swagger, we must define all api models for service.
This PR complete this too.

# Minor
- bugfix: some known minor issues in UI
- code clean.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1199,12 +1199,8 @@
       "description": "Operations related to server"
     },
     {
-      "name": "history",
-      "description": "Operations related to history"
-    },
-    {
-      "name": "models",
-      "description": "Operations related to models"
+      "name": "rules",
+      "description": "Operations related to rules"
     },
     {
       "name": "testing",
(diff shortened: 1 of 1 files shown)


internal/server/webui.go
```diff
@@ -941,43 +921,200 @@ func (s *Server) authMiddleware() gin.HandlerFunc {
...
+			if c.Request.Method == "OPTIONS" {
+				c.AbortWithStatus(204)
+				return
+			}
+			c.Next()
+		},
+	)
+
+	// Create authenticated API group
+	authAPI := manager.NewGroup("api", "v1", "")
```
FFengIll, line 967: key change here

PR #32, opened 2025-12-13
feat/stream: support to request openai and anthropic in stream mode.

FFengIll:
# Major: support stream mode
Stream is default mode for many application, so we need to implement it

done #33 

# Minor
- feat: redirect some path to no route 404
- feat: simplify log middleware

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1,11 +1,8 @@
 package server
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/server/anthropic.go
```diff
@@ -30,6 +32,26 @@ func (s *Server) AnthropicMessages(c *gin.Context) {
 		c.Request.Body = io.NopCloser(strings.NewReader(string(bodyBytes)))
 	}
 
+	// Parse the request to check if streaming is requested
```
FFengIll, line 35: stream option is special, we should do another parse.

PR #41, opened 2025-12-15
Feat: debug middleware

andreasfoo:
##  debug middleware
All requests and it's corresponding response would log into .tingly-box/log/bad_request.log for all http code >400 if --debug option is feed to tingly start.

## add a option --adaptor 
The openai/anstropic style openapi adaption swith, which is turned off by default, since we have some problem:
image

## make all tests for task go:test pass

### DIFF
--- a/internal/server/openai.go
+++ b/internal/server/openai.go
@@ -251,7 +251,7 @@ func (s *Server) convertAnthropicResponseToOpenAI(
 	responseModel string,
 ) map[string]interface{} {
 
-	var message map[string]interface{}
+	message := make(map[string]interface{})
 	var toolCalls []map[string]interface{}
 	var textContent string
 


internal/server/openai.go
```diff
@@ -250,20 +246,77 @@ func (s *Server) ListModels(c *gin.Context) {
 }
 
 // convertAnthropicResponseToOpenAI converts an Anthropic response to OpenAI format
-func (s *Server) convertAnthropicResponseToOpenAI(anthropicResp *AnthropicMessagesResponse, responseModel string) map[string]interface{} {
+func (s *Server) convertAnthropicResponseToOpenAI(
+	anthropicResp *anthropic.Message,
+	responseModel string,
+) map[string]interface{} {
+
+	message := map[string]interface{}{
```
FFengIll: why init as this?

internal/server/openai.go
```diff
@@ -250,20 +246,77 @@ func (s *Server) ListModels(c *gin.Context) {
 }
 
 // convertAnthropicResponseToOpenAI converts an Anthropic response to OpenAI format
-func (s *Server) convertAnthropicResponseToOpenAI(anthropicResp *AnthropicMessagesResponse, responseModel string) map[string]interface{} {
+func (s *Server) convertAnthropicResponseToOpenAI(
```
FFengIll, line 249: we should move all convertor into pkg (make them public)

internal/server/openai.go
```diff
@@ -273,66 +326,110 @@ func (s *Server) convertAnthropicResponseToOpenAI(anthropicResp *AnthropicMessag
...
 			}
+			continue
 		}
-	}
 
-	// Create Anthropic request parameters
-	params := anthropic.MessageNewParams{
-		Model:    anthropic.Model(model),
-		Messages: messages,
+		//3⃣ Tool result message → tool_result block (must be USER role)
```
FFengIll: pls remove special chars

internal/server/server.go
```diff
@@ -86,15 +94,25 @@ func NewServerWithOptions(cfg *config.Config, useUI bool) *Server {
...
-		useUI:      useUI,
-		assets:     assets,
+		config:        cfg,
+		jwtManager:    jwtManager,
+		router:        gin.New(),
+		logger:        memoryLogger,
+		useUI:         useUI,
+		assets:        assets,
 		clientPool: NewClientPool(), // Initialize client pool
+		debugMW:       debugMW,
```
FFengIll, line 114: should we set deubg mw here (as field) or option control with useXX?

internal/server/server.go
```diff
@@ -241,6 +264,11 @@ func (s *Server) setupRoutes() {
 	api := s.router.Group("/api")
 	api.Use(s.UserAuth()) // Require user authentication for management APIs
 	{
+		// Debug logging control endpoints
```
FFengIll: Suggest to run debug with setup options only, aka. DO NOT support dynamic switch.

PR #70, opened 2025-12-21
Refactor: extract auth middleware and update provider model with UUID.

FFengIll:
# Major: update provider model to use uuid as PK
This PR update frontend and backend both which migrate providers to a array with uuid.
All related code is updated.

# Minor
- refactor: extract auth middleware.
- test: update testing, and mention that we use name as uuid in testing for convenience.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -27,10 +27,8 @@ function App() {
                                         <Routes>
                                             <Route path="/" element={<Home />} />
                                             <Route path="/home" element={<Home />} />
-                                            <Route path="/dashboard" element={<Home />} />
-                                            <Route path="/credential" element={<ProviderPage />} />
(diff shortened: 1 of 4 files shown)


frontend/src/components/ModelSelectTab.tsx
```diff
@@ -38,32 +38,43 @@ export interface ProviderSelectTabOption {
...
 
+    // Create provider name to UUID mapping for search functionality
+    const providerNameToUuid = React.useMemo(() => {
+        const map: { [name: string]: string } = {};
+        providers.forEach(provider => {
+            map[provider.name] = provider.uuid;
+        });
+        return map;
+    }, [providers]);
+
```
andreasfoo, line 77: so we should keep name unique, maybe the backend should do this validation?

Issue #90, opened 2025-12-26
[feature] Maintain a provider config file in remote.


FFengIll:
We should maintain a provider config file in repo (as remote) to provide information about
- ai model provider name, type and base url
- models meta data for providers

then we can keep refresh in remote, and fallback to local (embed) too.

PR #94, opened 2025-12-26
minor: add a model template system at backend.

andreasfoo:
resolve[issue 90]( https://github.com/tingly-dev/tingly-box/issues/90)

### introduces a model template resolution system with a three-level fallback chain to ensure reliable model discovery for each provider.

Resolution Order:

1. Provider API
    Fetch the model list directly from the provider when available.

2. GitHub Repository
    Fall back to a maintained GitHub source if the provider API is unavailable or incomplete.

3. Embedded Resource File
    Use a built-in, embedded resource file (shipped with the binary) as the final fallback to guarantee availability at deploy time.

### DIFF
--- a/internal/server/tests/load_balancer_test.go
+++ b/internal/server/tests/load_balancer_test.go
@@ -875,7 +875,7 @@ func TestLoadBalancer_WeightedRandom(t *testing.T) {
 	// Test weighted selection (run multiple times to see distribution)
 	provider1Count := 0
 	provider2Count := 0
-	total := 100
+	total := 400
 
 	for i := 0; i < total; i++ {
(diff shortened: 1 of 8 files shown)


internal/server/tests/load_balancer_test.go
```diff
@@ -875,7 +875,7 @@ func TestLoadBalancer_WeightedRandom(t *testing.T) {
 	// Test weighted selection (run multiple times to see distribution)
 	provider1Count := 0
 	provider2Count := 0
-	total := 100
+	total := 400
```
FFengIll, line 878: why 400?

PR #97, opened 2025-12-27
feat: support adaptor from anthropic to openai request and openai to anthropic response.

FFengIll:
# Major: update adaptor from anthropic client to openai service
- feat: support stream response
- feat: support tool, tool call, tool result
- refactor: better naming
- test: add test for stream adaptor

# Minor
- bugfix in routes using adaptor


pkg/adaptor/request_to_openai.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	// Map stop reason
+	finishReason := "stop"
+	switch anthropicResp.StopReason {
+	case "tool_use":
+		finishReason = "tool_calls"
+	case "max_tokens":
+		finishReason = "length"
+	}
```
andreasfoo, line 72: Maybe we need a default to log?

FFengIll: default is `stop` for a to o.

Issue #92, opened 2025-12-26
[feature] add server option to control auto open


FFengIll:
Current version will auto open a web page for web-ui, please add some option control to make GUI disable this behavior.

PR #101, opened 2025-12-29
minor: add verbose and debug option in config, as well as a open brow…

andreasfoo:
resolve [issue](https://github.com/tingly-dev/tingly-box/issues/92)

### Move --debug and --verbose to config file

- debug and verbose can now be configured via the config file

- Priority order: CLI options > config file > default values

### Add --open-browser option

- Introduces a new open_browser option to control whether the main page URL is opened automatically

- The macOS app bundle sets this option to false to prevent auto-opening the main page on launch

### DIFF
--- a/internal/wails3/services/proxy_service.go
+++ b/internal/wails3/services/proxy_service.go
@@ -38,7 +38,7 @@ func NewUIService(configDir string, port int) (*ProxyService, error) {
 		appConfig,
 		manager.WithUI(true),
 		manager.WithAdaptor(false),
-		manager.WithDebug(true),
+		manager.WithDebug(false),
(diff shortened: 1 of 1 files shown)


internal/wails3/services/proxy_service.go
```diff
@@ -39,6 +39,7 @@ func NewUIService(configDir string, port int) (*ProxyService, error) {
 		manager.WithUI(true),
 		manager.WithAdaptor(false),
 		manager.WithDebug(true),
```
FFengIll: we may change this to false?

FFengIll: `debug`

PR #110, opened 2025-12-31
feat: make log middleware's filter configurable; add a cache mech for provider template

andreasfoo:
##  make log middleware's filter predicate configurable
see conf file-> error_log_filter_expression default: StatusCode >= 400 && Path matches '^/api/'
  | Goal                            | Expression                                   |
  |---------------------------------|----------------------------------------------|
  | Log everything                  | "true"                                       |
  | Log nothing                     | "false"                                      |
  | Only server errors (5xx)        | "StatusCode >= 500"                          |
  | Only client errors (4xx)        | "StatusCode >= 400 && StatusCode < 500"      |
  | Only POST errors                | "StatusCode >= 400 && Method == 'POST'"      |
  | API endpoint errors             | "StatusCode >= 400 && Path matches '^/api/'" |
  | Slow requests (>1s) with errors | "StatusCode >= 400 && Duration > 1000"       |
  | Specific status codes           | "StatusCode in [400, 401, 403, 404, 500]"    |

## add a cache mechanism for provider template
Fetching provider template at boot leads to significant start up performance problem. We add a cache mechanism with a 12h ttl.

### DIFF
--- a/internal/config/constants.go
+++ b/internal/config/constants.go
@@ -33,6 +33,10 @@ const (
 	DefaultTokenThreshold   = int64(10000) // Default token threshold for token-based and hybrid tactics
 
 	DefaultTemplateHTTPTimeout = 30 * time.Second // Default HTTP timeout for fetching templates
+
+	// Template cache constants
+	DefaultTemplateCacheTTL = 12 * time.Hour // Default TTL for template cache
+	TemplateCacheFileName   = "provider_template.json"
 )
 
 // GetTinglyConfDir returns the config directory path (default: ~/.tingly-box)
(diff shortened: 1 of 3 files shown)


internal/config/constants.go
```diff
@@ -33,6 +33,10 @@ const (
 	DefaultTokenThreshold   = int64(10000) // Default token threshold for token-based and hybrid tactics
 
 	DefaultTemplateHTTPTimeout = 30 * time.Second // Default HTTP timeout for fetching templates
+
+	// Template cache constants
+	DefaultTemplateCacheTTL = 24 * time.Hour // Default TTL for template cache
+	TemplateCacheFileName   = "template_cache.json"
```
FFengIll: require clear naming

PR #117, opened 2026-01-02
bugfix: missing fields process for o to a stream (especially, thinking); bugfix: missing token count for openai provider.

FFengIll:
# Major: New OpenAI to Anthropic stream adaptor
- support more fields, especially thinking.
- add some edge cases process, including empty chunk, reasoning chunk, block order.

# Minor: Forget to handle token count for openai style provider

### DIFF
--- a/pkg/adaptor/stream_to_anthropic_test.go
+++ b/pkg/adaptor/stream_to_anthropic_test.go
@@ -8,44 +8,45 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/anthropics/anthropic-sdk-go"
-	anthropicOption "github.com/anthropics/anthropic-sdk-go/option"
 	"github.com/gin-gonic/gin"
 	"github.com/openai/openai-go/v3"
+	openaiOption "github.com/openai/openai-go/v3/option"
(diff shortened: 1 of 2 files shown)


pkg/adaptor/stream_to_anthropic.go
```diff
@@ -265,13 +306,216 @@ func sendAnthropicStreamEvent(c *gin.Context, eventType string, eventData map[st
...
+func mergeMaps(base map[string]interface{}, extra map[string]interface{}) map[string]interface{} {
+	if extra == nil || len(extra) == 0 {
+		return base
+	}
+	if base == nil {
+		base = make(map[string]interface{})
+	}
+	for k, v := range extra {
+		base[k] = v
+	}
```
andreasfoo, line 335: Extra would overwrite base, is it by design?

andreasfoo (approved):

pkg/adaptor/stream_to_anthropic.go
FFengIll: since it comes from openai response, extras are not same as base; we may guard them in future.

PR #127, opened 2026-01-05
 Minor: "New API Key" Dialog update

andreasfoo:
Add New API Key Dialog Changes:

- Merge Verification: Remove the "Verify" button and merge the logic with the "Add API Key" action.

- Conditional Validation: If no API key is provided, skip the validation/verification step entirely.

- UI/UX Update: Replace the "Cancel" button with a close icon (X) in the top-right corner of the box. Clicking it should close the dialog.

- Keyless Support: Add a "No API Key" checkbox to the right of the API Key input field.
"No Key Required" are invisible if we choose from dropdown:
image

"No Key Required" are visible if we customize baseurl:
image

### DIFF
--- a/frontend/src/i18n/locales/en.ts
+++ b/frontend/src/i18n/locales/en.ts
@@ -26,7 +26,7 @@ export default {
     "slogan": "Ready for AI Agent Dev",
     "version": "version {{version}}",
     "nav": {
-      "home": "Home",
+      "home": "Scenario",
       "settings": "Settings",
       "useOpenAI": "OpenAI SDK",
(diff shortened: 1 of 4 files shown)


frontend/src/i18n/locales/en.ts
```diff
@@ -26,7 +26,7 @@ export default {
     "slogan": "Ready for AI Agent Dev",
     "version": "version {{version}}",
     "nav": {
-      "home": "Home",
+      "home": "Scenarios",
```
FFengIll: use singular

PR #152, opened 2026-01-07
feat: extract client define for llm and separate anthropic v1 and v1 beta adaptor.

FFengIll:
# Major: Separate support for anthropic v1 and v1 beta
- adaptor: support both v1 and v1 beta
- build: import submodule for anthropic-sdk-go fork (with bugfix)
- refactor: extract llm client and http client
- feat: use provider and model together as client key

# Minor: Update CLI with log file output init

### DIFF
--- a/internal/server/anthropic_v1.go
+++ b/internal/server/anthropic_v1.go
@@ -3,7 +3,6 @@ package server
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/server/anthropic_v1.go
```diff
@@ -0,0 +1,276 @@
...
+		// Use OpenAI conversion path (default behavior)
+		if isStreaming {
+			// Convert Anthropic request to OpenAI format for streaming
+			openaiReq := adaptor.ConvertAnthropicToOpenAIRequest(&req, true)
+
+			da, _ := json.MarshalIndent(req, "", "    ")
+			fmt.Printf("requst to openai: %s\n", da)
+
+			do, _ := json.MarshalIndent(openaiReq, "", "    ")
+			fmt.Printf("requst to openai: %s\n", do)
```
andreasfoo: maybe we print too much?

pkg/adaptor/stream_to_anthropic_beta.go
```diff
@@ -0,0 +1,396 @@
...
+				"type": deltaTypeTextDelta,
+				"text": delta.Content,
+			}
+			deltaMap = mergeMaps(deltaMap, currentExtras)
+			sendBetaContentBlockDelta(c, state.textBlockIndex, deltaMap, flusher)
+		} else if choice.FinishReason == "" && state.textBlockIndex != -1 {
+			// Send empty delta for empty chunks to keep client informed
+			// Only if text block has been initialized
+			currentExtras := parseRawJSON(delta.RawJSON())
+			currentExtras = filterSpecialFields(currentExtras)
```
andreasfoo, line 180: these fields we can not do transformation, so we should filter it?

FFengIll: yes, we have copy it into thinking block for anthropic, remove it is ok for O stream to A

PR #161, opened 2026-01-09
fix: Auto-fetch provider models when cache is empty

0xbeekeeper:
When opening the model select dialog, the model list was empty and required manual refresh. This fix:

1. Auto-fetches models from Provider API when local cache returns empty
2. Fetches first provider's models when dialog opens


FFengIll (approved):
Thanks.

PR #199, opened 2026-01-18
Fix Anthropic base path prefixes and add Xiaomi MiMo template

PawiX25:
## Problem
- Anthropic-style providers with path prefixes (e.g., https://api.xiaomimimo.com/anthropic) were forwarded to https://api.xiaomimimo.com/v1/... because the SDK built absolute /v1 paths, dropping the prefix.
- Provider probe OPTIONS for Anthropic used the raw base URL (without /v1), which caused false probe failures with some providers.

## Fix
- Normalize Anthropic base URLs by splitting host vs. path prefix and rewriting request paths to preserve the prefix.
- OPTIONS probe for Anthropic targets /v1.
- Add Xiaomi MiMo provider template with docs and base URLs.

## Testing
- Local UI load.
- Direct Xiaomi Anthropic endpoint returned 200 (manual curl).


internal/server/probe_handler.go
```diff
@@ -413,7 +413,16 @@ func (s *Server) probeOptionsEndpoint(provider *typ.Provider) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	req, err := http.NewRequestWithContext(ctx, "OPTIONS", provider.APIBase, nil)
+	optionsURL := provider.APIBase
```
FFengIll, line 416: Yep, you are right here

internal/llmclient/anthropic.go
```diff
@@ -23,13 +24,71 @@ type AnthropicClient struct {
...
+
 // defaultNewAnthropicClient creates a new Anthropic client wrapper
 func defaultNewAnthropicClient(provider *typ.Provider) (*AnthropicClient, error) {
-	// Handle API base URL - Anthropic SDK expects base without /v1
-	apiBase := provider.APIBase
-	if strings.HasSuffix(apiBase, "/v1") {
-		apiBase = apiBase[:len(apiBase)-3]
-	}
+	// Handle API base URL - Anthropic SDK expects base without /v1 and without extra path prefixes
+	apiBase, pathPrefix := normalizeAnthropicBase(provider.APIBase)
```
FFengIll: Discuss: since llmclient is used only for some special cases like probe, this `pathPrefixRoundTripper` maybe too fabric.

Would you mind just try (below) here?
```
	optionsURL := provider.APIBase 1Code has comments. Press enter to view.
	if provider.APIStyle == typ.APIStyleAnthropic {
		apiBase := strings.TrimSuffix(provider.APIBase, "/")
		if !strings.Contains(apiBase, "/v1") {
			apiBase = apiBase + "/v1"
		}
		optionsURL = apiBase
	}
```

Furthermore, tingly-box request model provider via a robost implement which will process anthropic and openai "v1" or "beta" internal (ref ./libs)

FFengIll (changes requested):

FFengIll:
> I tried the dual `OPTIONS` check (both raw `APIBase` and `APIBase` + `/v1`) but still received a `404` from the Xiaomi MiMo endpoint. It appears `OPTIONS` isn’t supported there.
> 
> Would you be open to treating `404`/`405` responses as non-fatal—or skipping `OPTIONS` for Anthropic-style bases—to avoid false negatives? If not, I’ll leave the probe as-is and focus on other issues.

Thanks for investigating the issue; I’ve identified the root cause. 

I suggest handling the Mi Provider in the current PR by implementing a small client-side modification (removing the RoundTripper). 

We will soon support new frontend interaction logic that allows users to continue adding an API key even if the automatic probe fails, preventing potential edge cases.

internal/config/template/provider_templates.json
```diff
@@ -547,6 +547,23 @@
...
+      "website": "https://platform.xiaomimimo.com",
+      "description": "Xiaomi MiMo API Open Platform",
+      "type": "official",
+      "api_doc": "https://platform.xiaomimimo.com/#/docs/api/text-generation/openai-api",
+      "model_doc": "https://platform.xiaomimimo.com/#/docs/pricing",
+      "pricing_doc": "https://platform.xiaomimimo.com/#/docs/pricing",
+      "base_url_openai": "https://api.xiaomimimo.com/v1",
+      "base_url_anthropic": "https://api.xiaomimimo.com/anthropic",
+      "models": [
+        "mimo-v2-flash"
```
FFengIll, line 564: Thanks for the contribute

PR #228, opened 2026-01-21
Add a pass through mode ; Add missing usage field from response.

andreasfoo:
This PR introduces the following improvements:

##  New Pass-Through Mode

- Adds a --pass-through flag to the CLI to enable a pass-through mode.

- When enabled, this bypasses the usual request/response transformations, which makes it easier to debug raw API traffic.

- This mode is integrated into the server manager and CLI options so it can be toggled at startup.

test with:
```bash
curl -s http://127.0.0.1:12580/passthrough/openai/v1/chat/completions \
 -X POST \
-H 'Content-Type: application/json' -H 'Authorization: Bearer tingly-box-xxx'        \
-d '{"model":"qwen-plus","messages":[{"role":"user","content":"Hi"}],"stream":false}'



curl -s http://127.0.0.1:12580/passthrough/anthropic/messages \
        -X POST \
        -H 'Content-Type: application/json' \
        -H "Authorization: Bearer tingly-box-xxx" \
        -d '{"model":"tingly/anthropic","messages":[{"role":"user","content":"Hi"}],"max_tokens":100}'
```
## Preserve server_tool_use Field in Transformed Responses

Fixes the response transformation logic to retain the server_tool_use field when adapting responses between formats.


internal/server/middleware/passthrough_middleware.go
```diff
@@ -0,0 +1,337 @@
+package middleware
```
FFengIll: How about to add new endpoint `/passthrough/openai` & `/passthrough/anthropic`.
Then we do not require cli option, mw, and so on, just use these two for passthrough.

internal/server/passthrough.go
```diff
@@ -0,0 +1,331 @@
...
+
+	// Copy response headers
+	for key, values := range resp.Header {
+		for _, value := range values {
+			c.Header(key, value)
+		}
+	}
+
+	// Handle streaming vs non-streaming
+	if resp.Header.Get("Content-Type") == "text/event-stream" {
```
FFengIll: Use contains here, since prefix changable.

internal/server/server.go
```diff
@@ -397,6 +397,21 @@ func (s *Server) UseAIEndpoints() {
...
+	s.SetupPassthroughOpenAIEndpoints(passthroughOpenai)
+
+	passthroughAnthropic := s.engine.Group("/passthrough/anthropic")
+	s.SetupPassthroughAnthropicEndpoints(passthroughAnthropic)
+
+	// Versioned passthrough routes
+	passthroughOpenaiV1 := s.engine.Group("/passthrough/openai/v1")
+	s.SetupPassthroughOpenAIEndpoints(passthroughOpenaiV1)
+
+	passthroughAnthropicV1 := s.engine.Group("/passthrough/anthropic/v1")
```
FFengIll: only require /v1 version
but do not set v1 for anthrpoic client sdk.

internal/server/passthrough.go
```diff
@@ -0,0 +1,324 @@
...
+
+	// Copy response headers
+	for key, values := range resp.Header {
+		for _, value := range values {
+			c.Header(key, value)
+		}
+	}
+
+	// Handle streaming vs non-streaming
+	if strings.Contains(resp.Header.Get("Content-Type"), "text/event-stream") {
```
FFengIll, line 171: Use contains here, since prefix changable.

internal/server/server.go
```diff
@@ -397,6 +397,18 @@ func (s *Server) UseAIEndpoints() {
...
+	s.SetupPassthroughOpenAIEndpoints(passthroughOpenai)
+
+	passthroughAnthropic := s.engine.Group("/passthrough/anthropic")
+	s.SetupPassthroughAnthropicEndpoints(passthroughAnthropic)
+
+	// Versioned passthrough routes
+	passthroughOpenaiV1 := s.engine.Group("/passthrough/openai/v1")
+	s.SetupPassthroughOpenAIEndpoints(passthroughOpenaiV1)
+
 	// scenario
```
FFengIll, line 412: only require /v1 version
but do not set v1 for anthrpoic client sdk.

PR #248, opened 2026-01-26
gui: refresh gui implementation with command and service entry refactoring.

FFengIll:
This PR provide new gui implementation with some related refactory and bugfix.
Furthermore, this PR separate gui version into individual module to reduce tingly-box deps.

### DIFF
--- a/build/darwin/Info.plist
+++ b/build/darwin/Info.plist
@@ -14,7 +14,7 @@
         <key>CFBundleGetInfoString</key>
             <string>This is a comment</string>
         <key>CFBundleShortVersionString</key>
-            <string>0.1.0</string>
+            <string>0.1</string>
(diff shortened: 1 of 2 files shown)


build/darwin/Info.plist
```diff
@@ -14,29 +14,25 @@
...
+                <key>NSAllowsLocalNetworking</key>
+                <true/>
             </dict>
-        </dict>
+        <key>NSAppSandboxEnabled</key>
+            <true/>
+        <key>LSApplicationCategoryType</key>
+            <string>public.app-category.developer-tools</string>
+        <key>ITSAppUsesNonExemptEncryption</key>
+            <true>
```
andreasfoo: whats the true tag mean, seems wrong?

PR #256, opened 2026-01-27
feat: add apply configuration button for open code and claude code.

andreasfoo:
## Changes:

- Added “Apply Configuration” button to Open Code and Claude Code panels.

- Users can now apply configuration changes directly, improving visibility and user flow.

## FrontEnd preview
image

### DIFF
--- a/frontend/src/services/api.ts
+++ b/frontend/src/services/api.ts
@@ -760,7 +760,7 @@ export const api = {
     },
 
     // Config Apply API - Safe endpoints that generate config from system state
-    applyClaudeConfig: async (mode: string = 'unified'): Promise<any> => {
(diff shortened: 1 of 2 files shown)


frontend/src/services/api.ts
```diff
@@ -758,6 +758,27 @@ export const api = {
             return { success: false, error: error.message };
         }
     },
+
+    // Config Apply API - Safe endpoints that generate config from system state
+    applyClaudeConfig: async (mode: string = 'unified'): Promise<any> => {
```
FFengIll: DO NOT set default value here.

andreasfoo: It's always called with args, but ok, as you wish

PR #267, opened 2026-01-30
fix: v1 anthropic interface also supports Response.

andreasfoo:
Enable OpenCode  v1 requests with beta=false to correctly produce a v1 response. The previous implementation only handled the beta (v1beta) flow and did not cover the v1 → response path.

## Context

For incoming requests, we support two flows:

- v1 endpoint (beta=false) → v1 types → Responses API → v1 response

- v1beta endpoint (beta=true) → v1beta types → Responses API → v1beta response

Previously, the response-building logic effectively only worked for the beta flow, so v1 requests were missing the proper v1 response handling.

## Implementation options considered

There are two ways to implement this:

### Option 1 (separate v1 path): Implement the missing v1 → v1 response handling directly.

- Pros: clear version separation, easier to reason about

- Cons: some logic overlaps with beta

### Option 2 (bridge/merge via beta path): Map v1 into the beta response path and reuse beta logic.

- Pros: potentially less duplication

- Cons: ends up being hard to follow and mixes version boundaries

I initially implemented Option 2, but the logic became confusing/fragile. I switched back to Option 1 for clarity and maintainability, at the cost of some duplicated logic with beta.


internal/loadbalance/load_balancing_test.go
```diff
@@ -1,14 +1,26 @@
-package loadbalance
+package loadbalance_test
```
FFengIll: should be loadbalance too

internal/server/anthropic_v1_beta.go
```diff
@@ -505,9 +505,6 @@ func (s *Server) handleAnthropicV1BetaViaResponsesAPIStreaming(c *gin.Context, r
 		recorder = r.(*ScenarioRecorder)
 	}
 	streamRec := newStreamRecorder(recorder)
-	if streamRec != nil {
```
FFengIll, line 508: we require this for recording.

internal/protocol/stream/stream_openai_to_anthropic_beta.go
FFengIll, line 1: why do we move from this file to non-beta file?

FFengIll (changes requested):

internal/protocol/stream/stream_openai_to_anthropic_beta.go
andreasfoo: We didn’t intend to move it. This isn’t just a relocation—it’s more of an abstraction. We now pass sender functions so it works in both beta and non-beta scenarios. But ok, move back to beta may be more convenient to review.

PR #286, opened 2026-02-03
refactor: code clean and better naming without logic change.

FFengIll:
Enable adaptor as default and remove related check logic.
Better naming for files and move some definitions.
Separate rule and service select logic.

### DIFF
--- a/internal/server/openai_responses.go
+++ b/internal/server/openai_responses.go
@@ -69,8 +69,6 @@ func (s *Server) ResponsesCreate(c *gin.Context) {
 		return
 	}
 
-	responseModel := string(req.Model)
-
 	// Determine provider & model
 	var (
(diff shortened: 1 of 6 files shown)


internal/server/openai_responses.go
```diff
@@ -88,7 +86,19 @@ func (s *Server) ResponsesCreate(c *gin.Context) {
 		})
 		return
 	}
-	provider, selectedService, rule, err = s.DetermineProviderAndModelWithScenario(scenarioType, responseModel, nil)
```
FFengIll, line 91: here is a bug since we do not support data extract for responses in smart routing mode.

FFengIll: we may fix it in another PR.

PR #306, opened 2026-02-06
bugfix: openai stream data structure error and anthropic to openai stream context error

FFengIll:
Bugs are fixed in this PR.
- openai stream SSE `special space` missing, e.g. `data: [DONE]` (take care of the space).
- anthropic to openai stream context error - context cancel too early.

Both of them bother xcode intelligence.

### DIFF
--- a/internal/server/openai_chat.go
+++ b/internal/server/openai_chat.go
@@ -216,10 +216,12 @@ func (s *Server) handleOpenAIStreamResponse(c *gin.Context, stream *ssestream.St
 	}()
 
 	// Set SSE headers (mimicking OpenAI response headers)
-	c.Header("Access-Control-Allow-Headers", "Origin, Content-Type, Authorization")
+	c.Header("Access-Control-Allow-Headers", "Origin, Content-Type, Authorization, Cache-Control")
 	c.Header("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS")
 	c.Header("Access-Control-Allow-Origin", "*")
 	c.Header("Content-Type", "text/event-stream; charset=utf-8")
+	c.Header("Cache-Control", "no-cache")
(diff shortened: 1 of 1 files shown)


internal/protocol/stream/anthropic_to_openai.go
```diff
@@ -163,7 +164,8 @@ func HandleAnthropicToOpenAIStreamResponse(c *gin.Context, req *anthropic.Messag
 
 			sendOpenAIStreamChunk(c, chunk)
 			// Send final [DONE] message
-			c.SSEvent("", "[DONE]")
+			// MENTION: must keep extra space (matching openai_chat.go:462)
+			c.SSEvent("", " [DONE]")
```
andreasfoo, line 168: OMG, terrible logic...

internal/server/openai_chat.go
```diff
@@ -199,12 +215,11 @@ func (s *Server) handleOpenAIStreamResponse(c *gin.Context, stream *ssestream.St
...
-	// Set SSE headers
-	c.Header("Content-Type", "text/event-stream")
-	c.Header("Cache-Control", "no-cache")
-	c.Header("Connection", "keep-alive")
+	// Set SSE headers (mimicking OpenAI response headers)
+	c.Header("Access-Control-Allow-Headers", "Origin, Content-Type, Authorization")
+	c.Header("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS")
 	c.Header("Access-Control-Allow-Origin", "*")
-	c.Header("Access-Control-Allow-Headers", "Cache-Control")
+	c.Header("Content-Type", "text/event-stream; charset=utf-8")
```
andreasfoo, line 225: why we need these headers and remove keep-alive?

FFengIll: most provider share this, so we keep this.

Issue #251, opened 2026-01-26
Value of "Provider or Custom Base URL" is odd after double select the same item in dropdown list


KrabsWong:
## Issue happened in panel "/api-keys" when select "provider"
<img width="646" height="832" alt="Image" src="https://github.com/user-attachments/assets/dde686c0-d957-4164-856e-97fd91c21eaa" />

As the screenshot above, if you select the same item twice, then the data in the Input field is odd

### The way to reproduce the issue

First click: Link only
<img width="581" height="124" alt="Image" src="https://github.com/user-attachments/assets/24868be4-2cf8-4c1f-9381-819a50e24dfe" />

Second click: Link with prefix
<img width="578" height="121" alt="Image" src="https://github.com/user-attachments/assets/b1c72246-414c-4099-8033-69ca5471845e" />

Click "Add api key", error happened
<img width="608" height="664" alt="Image" src="https://github.com/user-attachments/assets/1b5d0220-2c1f-4769-8d71-afbea645f4d4" />

PR #309, opened 2026-02-06
bugfix: improve provider autocomplete selection handling

KrabsWong:
For fixing this issue https://github.com/tingly-dev/tingly-box/issues/251

  - Fix API base URL extraction when user selects from dropdown
  - Add custom filter to show all options when input matches selected format
  - Display "Title - URL" format in input when provider is selected
  - Only update apiBase on actual user input, not on selection


FFengIll (approved):
Thanks for your contribution.

Issue #323, opened 2026-02-09
[BUG] Using local vllm serve with Tingly Box


2295181208-cloud:
BackGround：I using Claude Code v2.1.37， vllm 0.15.0 with local GLM-4.7 model，Tingly Box
我是将vllm serve部署在远端服务器上，然后本地macos使用的claude code
这个是我的claude code，我让他测试一下本地的工具直接中断，查看Tingly Box日志正常请求到了，查看vllm serve日志也看不出来任何的问题
<img width="1515" height="70" alt="Image" src="https://github.com/user-attachments/assets/2304e51b-e92d-41d4-a942-78c7ff75191c" />
<img width="1776" height="982" alt="Image" src="https://github.com/user-attachments/assets/d0f38649-d23d-4723-a515-0e19cac748ff" />
<img width="983" height="473" alt="Image" src="https://github.com/user-attachments/assets/934b9054-dd48-4854-844d-a7155154811e" />
<img width="2546" height="1500" alt="Image" src="https://github.com/user-attachments/assets/f6c76419-fcd9-4fd6-adf6-592bbdeb1133" />

然后当我使用vllm serve OpenAI-Style API KEY的时候
vllm serve会报错找不到function，claude code中所有的工具调用会用代码的形式输出出来
<img width="2536" height="1458" alt="Image" src="https://github.com/user-attachments/assets/76712981-faf1-4b25-bac3-91ca2676dc83" />
<img width="1774" height="208" alt="Image" src="https://github.com/user-attachments/assets/c008a835-f26b-4a18-bc7d-1b5f54213247" />

上面这些问题我在使用ccr的时候都没有遇到过，因为我在使用ccr的时候遇到了一个致命的问题，就是auto-compact无法使用，所以切换到这个router上，希望大佬能够帮我解答一下疑惑，感谢大佬
@FFengIll

FFengIll:
@2295181208-cloud Got it. I’ll look into the details over the next few days.

2295181208-cloud:
@FFengIll Hi！I’ve been running CCR + vLLM Serve + Claude Code locally for quite a while. Here are my vLLM Serve logs for reference. CCR is calling the /v1/chat/completions endpoint, which in our Tingly-Dev Project corresponds to the OpenAI protocol style. So far, everything has been working smoothly—both Claude Code’s tool-calling and the Claude Code for VS Code plugin features work without any issues on this stack.

[...]

PR #330, opened 2026-02-10
bugfix: tool choice error and always prefer reposnes api cause some provider error.

FFengIll:
This PR fix bugs in tool choice and openai/anthropic stream adaptor.
Root cause
1. BUG: toolchoice error in many protocol handler. 
    SOLUTION: check toolchoice and tools together.
2. BUG: always prefer to use responses api but implement with edge case error.
    SOLUTION: downgrade to chat api (except codex), and fix repsonses api later.
3. MINOR BUG: some index wrong in stream adaptor (but DO NOT cause error in most cases)

Feedback in #323 and fixed via vllm with Qwen3.

### DIFF
--- a/internal/protocol/request/openai_to_anthropic.go
+++ b/internal/protocol/request/openai_to_anthropic.go
@@ -123,10 +123,7 @@ func ConvertOpenAIToAnthropicRequest(req *openai.ChatCompletionNewParams, defaul
 		params.Tools = ConvertOpenAIToAnthropicTools(req.Tools)
 		// Convert tool choice
 		// ToolChoice is a Union type, check if any field is set
-		if req.ToolChoice.OfAuto.Value != "" || req.ToolChoice.OfAllowedTools != nil ||
-			req.ToolChoice.OfFunctionToolChoice != nil || req.ToolChoice.OfCustomToolChoice != nil {
-			params.ToolChoice = ConvertOpenAIToAnthropicToolChoice(&req.ToolChoice)
-		}
+		params.ToolChoice = ConvertOpenAIToAnthropicToolChoice(&req.ToolChoice)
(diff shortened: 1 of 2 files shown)


internal/protocol/request/openai_to_anthropic.go
```diff
@@ -121,13 +121,12 @@ func ConvertOpenAIToAnthropicRequest(req *openai.ChatCompletionNewParams, defaul
...
-	// ToolChoice is a Union type, check if any field is set
-	if req.ToolChoice.OfAuto.Value != "" || req.ToolChoice.OfAllowedTools != nil ||
-		req.ToolChoice.OfFunctionToolChoice != nil || req.ToolChoice.OfCustomToolChoice != nil {
-		params.ToolChoice = ConvertOpenAIToAnthropicToolChoice(&req.ToolChoice)
+		// Convert tool choice
+		// ToolChoice is a Union type, check if any field is set
+		if req.ToolChoice.OfAuto.Value != "" || req.ToolChoice.OfAllowedTools != nil ||
+			req.ToolChoice.OfFunctionToolChoice != nil || req.ToolChoice.OfCustomToolChoice != nil {
+			params.ToolChoice = ConvertOpenAIToAnthropicToolChoice(&req.ToolChoice)
+		}
```
andreasfoo: this condition is a little odd, why req.ToolChoice.OfAllowedTools != nil then we should convert tool choice? maybe (a || b) && (c|| d) ?

FFengIll: I changed this.
No need to add this condition, do everything in `ConvertOpenAIToAnthropicToolChoice` now

andreasfoo (approved):

PR #331, opened 2026-02-11
bugfix: openai responses api related protocol fields error.

FFengIll:
This PR fix bugs in openai responses api.

Bug details:
- BUG: function call fields are incomplete.
- BUG: some input field missing after marshal.
- BUG: should omit empty in stream response.

Feedback from #323


andreasfoo:
LGTM

PR #352, opened 2026-02-14
feat: model guardrails

seviezhou:
# Guardrails: initial policy management, built-ins, protected credentials, and credential masking

This PR introduces the first version of Guardrails, focused on making policy management and credential protection usable end to end.

### Policy and group management
- Introduces a policy-first Guardrails experience
- Adds clear policy categories:
  - Resource Access
  - Command Execution
  - Content
- Adds UI for creating, editing, and organizing policies
- Adds group management for shared defaults and policy organization
- Shows group-level enable/disable state in the policy list

### Built-in policies
- Adds a Builtins page for starting from predefined Guardrails policies
- Supports browsing and filtering built-in policies
- Installing a built-in opens the normal policy editor as a draft, and only saves after confirmation

### Protected credentials
- Adds a dedicated Protected Credentials page under Guardrails
- Allows users to create and manage protected credentials with generated alias tokens
- Keeps real secrets hidden in the UI by default
- Supports importing entries from the existing Credentials page
- Adds bulk selection and bulk delete flows

### Credential masking
- Adds credential masking so protected credentials can be replaced with aliases before content is sent to the model
- Supports creating mask policies directly from protected credentials
- Restores aliased credentials locally when needed for execution flows
- Improves visibility into masking activity in Guardrails history and overview

### History and overview
- Adds a Guardrails overview page with policy and activity summaries
- Adds better interception and masking history presentation
- Shows user-friendly credential information in history without exposing secrets

## Why this matters

This PR establishes the first usable Guardrails workflow:

- Users can define and organize protection policies
- Common protections are easier to adopt through built-ins
- Credentials can be protected from direct model exposure
- Guardrails activity becomes visible through overview and history pages

## User-facing outcome

With this PR, users can:

- Create and manage Guardrails policies and groups
- Start from built-in Guardrails policies
- Protect sensitive credentials with alias-based masking
- Generate mask policies from protected credentials
- Review Guardrails behavior through overview and history

## Notes

This PR introduces the initial Guardrails feature set, with emphasis on usability, policy management, and safe credential handling.

## Screenshot
截屏2026-03-31 10 22 59
截屏2026-03-31 10 23 08
截屏2026-03-31 10 23 13
截屏2026-03-31 10 23 23
截屏2026-03-31 10 23 30


internal/server/guardrails_credentials_handler.go
```diff
@@ -0,0 +1,258 @@
...
+		Enabled:     credential.Enabled,
+		SecretMask:  guardrails.MaskedSecret(credential.Secret),
+		CreatedAt:   credential.CreatedAt.Format("2006-01-02T15:04:05Z07:00"),
+		UpdatedAt:   credential.UpdatedAt.Format("2006-01-02T15:04:05Z07:00"),
+	}
+}
+
+// GetGuardrailsCredentials returns protected credentials without exposing raw secrets.
+func (s *Server) GetGuardrailsCredentials(c *gin.Context) {
+	store, err := s.guardrailsCredentialStore()
```
richardxx, line 82: 🔴 **CRITICAL**: This endpoint returns the raw secret in plaintext in the API response. This is a security vulnerability because:

1. API logs may capture these responses
2. Browser extensions could access the secret
3. Any XSS vulnerability would expose all credentials
4. Frontend may accidentally log or expose the secret

**Recommendation**: Never return raw secrets via API. Consider:
- Removing the Secret field from the response entirely
- Implementing a separate verification endpoint that requires re-authentication
- Using a reveal flow that requires explicit user confirmation

internal/guardrails/credential_store.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+		resolved = append(resolved, credential)
+	}
+	return resolved, nil
+}
+
+func (s *ProtectedCredentialStore) load() ([]ProtectedCredential, error) {
+	if s.path == "" {
+		return nil, fmt.Errorf("protected credential store path is empty")
+	}
```
richardxx, line 201: 🔴 **CRITICAL**: Credentials are stored in plaintext JSON with only file permissions (0600). This is a security vulnerability because:

1. Anyone with read access to the file can read all credentials
2. No protection against filesystem backup exposure
3. Temporary file created with same plaintext content
4. No integrity protection - file can be modified without detection

**Recommendation**: Implement encryption at rest:
- Use AES-256-GCM for authenticated encryption
- Derive encryption key from user passphrase or system keyring
- Consider using [age](https://github.com/FiloSottile/age) or [secretbox](https://pkg.go.dev/golang.org/x/crypto/nacl/secretbox)

internal/guardrails/credential.go
```diff
@@ -0,0 +1,116 @@
...
+		return "", fmt.Errorf("generate alias token: %w", err)
+	}
+	return AliasTokenPrefix + strings.ToUpper(hex.EncodeToString(buf)), nil
+}
+
+func normalizeStringSlice(values []string) []string {
+	if len(values) == 0 {
+		return nil
+	}
+	out := make([]string, 0, len(values))
```
richardxx, line 93: 🔴 **HIGH**: Only 8 bytes (64 bits) of randomness for alias token generation. This is a security concern because:

1. Total search space is ~1.8e19 possible tokens
2. With known prefix `TINGLY_SECRET_`, tokens are predictable in format
3. No rate limiting on credential operations to prevent brute force
4. An attacker who knows one alias token could potentially guess others

**Recommendation**: 
- Increase to at least 16 bytes (128 bits) of randomness
- Add rate limiting on credential endpoints
- Consider adding a timestamp or counter component

internal/server/guardrails_credentials_runtime.go
```diff
@@ -0,0 +1,419 @@
...
+		verdict := policy.Verdict
+		if verdict == "" && hasGroup {
+			verdict = group.DefaultVerdict
+		}
+		if verdict != guardrails.VerdictMask {
+			continue
+		}
+		scenarios := policy.Scope.Scenarios
+		if len(scenarios) == 0 && hasGroup {
+			scenarios = group.DefaultScope.Scenarios
```
richardxx: 🔴 **HIGH**: Potential race condition in credential mask state management. The `CredentialMaskState` is accessed and modified by concurrent goroutines during stream processing without proper synchronization.

The `maskData`, `masked`, `restored`, and `aliases` maps are all mutable state that could be accessed simultaneously from multiple goroutines processing stream events.

**Recommendation**: Add `sync.Mutex` protection to `CredentialMaskState` or make it immutable and return new instances on each modification.

internal/guardrails/credential_alias.go
```diff
@@ -0,0 +1,250 @@
...
+		AliasToReal: make(map[string]string),
+		RealToAlias: make(map[string]string),
+	}
+}
+
+func MayContainAliasToken(text string) bool {
+	return text != "" && strings.Contains(text, AliasTokenPrefix)
+}
+
+func (s *CredentialMaskState) remember(credential ProtectedCredential) {
```
richardxx, line 31: 🟡 **Code Quality**: This function silently fails without returning an error or logging. This makes debugging difficult and could lead to subtle bugs where credential masking doesn't work as expected.

**Recommendation**: Either return an error or log a warning when validation fails.

internal/guardrails/policy_content.go
```diff
@@ -0,0 +1,195 @@
...
+	return PolicyTypeContent
+}
+
+// Enabled returns whether the policy is enabled.
+func (r *ContentPolicy) Enabled() bool {
+	return r.enabled
+}
+
+// Evaluate matches text against configured patterns.
+func (r *ContentPolicy) Evaluate(_ context.Context, input Input) (PolicyResult, error) {
```
richardxx, line 106: 🟡 **Code Quality**: No validation of regex pattern complexity. A malicious admin could create a ReDoS (Regular Expression Denial of Service) vulnerability by providing a catastrophic backtracking pattern.

**Recommendation**: Add regex complexity validation and consider using a timeout for pattern matching.

internal/guardrails/credential_store.go
seviezhou: use sqlite db, the same as current Credentials.

internal/guardrails/credential.go
seviezhou: changed to 128bit

internal/guardrails/policy_content.go
seviezhou: already been checked inside `tingly-box/internal/guardrails/policy_content.go:55`

internal/server/guardrails_credentials_handler.go
seviezhou: need to show secret in frontend, the same logic as the Credentials page.

PR #363, opened 2026-02-22
bugfix: side bar is not auto refreshed after enable features

seviezhou:
## Summary

- Side bar is not auto refreashed after switching Global Experimental Features.
- Add `refresh()` method call after success `setScenarioFlag` calls.


FFengIll (approved):
LGTM

PR #367, opened 2026-02-24
feat: Service Health Check for Load Balancing 

andreasfoo:
### Service Health Check for Load Balancing

**Problem:**
When services, such as providers, are rate-limited (HTTP 429) or experience other errors (e.g., 401/403), the current round-robin load balancing mechanism may continue sending requests to them, leading to user-facing failures. There was no mechanism in place to detect unhealthy services and temporarily exclude them from the load balancing pool.

**Solution:**
This PR introduces a service health monitoring system, leveraging an efficient Filter + Picker architecture, which ensures that unhealthy services are excluded from the load balancing pool until they are deemed healthy again.

Key changes include:

1. **Health Monitor**: Tracks and updates the health status of services automatically based on various failure conditions.
2. **Health Filter**: Filters out unhealthy services before they are selected for load balancing.
3. **Probe-based Recovery**: Reuses the existing probe infrastructure to verify service health before marking a service as healthy again after recovery.

**How It Works:**

1. Requests are first passed through a **Health Filter** that checks the health of services.
2. The **Load Balancer Tactic** selects the next healthy service.
3. The **Health Monitor** reports on service health by tracking HTTP errors and using probes for recovery.

**Health State Transitions:**

| **Event**                        | **Action**                                 |
|----------------------------------|--------------------------------------------|
| 429 Rate Limit                   | Mark unhealthy immediately                 |
| 401/403 Auth Error               | Mark unhealthy immediately                 |
| 3 Consecutive Errors             | Mark unhealthy                             |
| Any Success                      | Mark healthy immediately                   |
| Recovery Timeout + Probe Success | Mark healthy                               |
| Recovery Timeout + Probe Fail    | Extend timeout                             |

**Key Features:**

- **Implicit Detection**: Health issues are detected through HTTP responses without requiring any special quota configuration.
- **Probe-based Recovery**: Services are probed before being marked as healthy again after a timeout, ensuring that recovery is validated.
- **Reuses Existing Infrastructure**: Utilizes the `AdaptiveProbe` system for health checks, ensuring consistent behavior across different service endpoints.
- **Configurable**: The health monitor is highly configurable with options for the recovery timeout, consecutive error thresholds, and enabling/disabling probes.

**Configuration:**
The following is set as default setting of health_monitor:
```yaml
health_monitor:
  consecutive_error_threshold: 3    # Mark service as unhealthy after N errors
  recovery_timeout_seconds: 300     # Auto-recovery timeout (5 minutes)
  probe_enabled: true               # Enable probe checks before recovery
```

### DIFF
--- a/internal/loadbalance/health_monitor.go
+++ b/internal/loadbalance/health_monitor.go
@@ -1,0 +1,305 @@
+package loadbalance
+
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+// HealthStatus represents the health state of a service
+type HealthStatus int
+
+const (
+	HealthHealthy HealthStatus = iota // Service is healthy and available
+	HealthUnhealthy                    // Service is unhealthy (rate limited, failing)
+)
+
+// String returns the string representation of HealthStatus
+func (h HealthStatus) String() string {
+	switch h {
+	case HealthHealthy:
+		return "healthy"
+	case HealthUnhealthy:
+		return "unhealthy"
+	default:
+		return "unknown"
+	}
+}
+
+// ServiceHealth tracks health information for a single service
+type ServiceHealth struct {
+	ServiceID         string        // Unique service identifier (provider:model)
+	Status            HealthStatus  // Current health status
+	LastError         error         // Last error that caused unhealthy state
+	LastErrorTime     time.Time     // When the error occurred
+	ConsecutiveErrors int           // Count of consecutive errors
+	RateLimited       bool          // True if last error was 429
+	AuthError         bool          // True if last error was 401/403
+	LastHealthCheck   time.Time     // Last time health was checked
+	RecoveryTimeout   time.Duration // Time before auto-recovery
+	mutex             sync.RWMutex  // Thread safety
+}
+
+// HealthProbeFunc is the function type for probing service health
+// Returns true if service is healthy, false otherwise
+type HealthProbeFunc func(serviceID string) bool
+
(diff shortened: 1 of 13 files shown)


internal/server/load_balancer.go
```diff
@@ -70,20 +72,44 @@ func (lb *LoadBalancer) SelectService(rule *typ.Rule) (*loadbalance.Service, err
 		return nil, fmt.Errorf("no active services for rule %s", rule.RequestModel)
 	}
 
-	// For single service rules, return it directly
-	if len(activeServices) == 1 {
-		return activeServices[0], nil
+	// Filter healthy services using health filter
+	healthyServices := lb.healthFilter.Filter(activeServices)
```
FFengIll, line 76: will it make request delay since health check cost some time.

internal/loadbalance/health_monitor.go
FFengIll, line 1: probe and health check might be a low level module in future, lb will use them but not control them.

FFengIll: we may change in future.

internal/server/load_balancer.go
andreasfoo: not too much, just a iteration of the rule's services.

internal/loadbalance/health_monitor.go
andreasfoo: Right. LB is now using probe to detect service liveness. Heath checking, for now is bind to LB, but should be open to others.

PR #375, opened 2026-02-25
migrate: define new db table for provider credentials and use them with migration.

FFengIll:
This PR implements a new database storage layer for persisting provider credentials, along with the corresponding migration.

- New database table for provider credentials
- Decoupled tool configuration from providers
- New database table for tool configuration in the General model
- New generalized architecture enabling support for both provider-specific and global tool configurations

### DIFF
--- a/internal/server/config/config.go
+++ b/internal/server/config/config.go
@@ -53,11 +53,7 @@ type Config struct {
 	DefaultMaxTokens int  `json:"default_max_tokens"` // Default max_tokens for anthropic API requests
 	Verbose          bool `json:"verbose"`            // Verbose mode for detailed logging
 	Debug            bool `json:"-"`                  // Debug mode for Gin debug level logging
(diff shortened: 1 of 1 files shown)


internal/data/db/provider_store.go
```diff
@@ -0,0 +1,588 @@
...
+	"gorm.io/gorm/logger"
+
+	"github.com/tingly-dev/tingly-box/internal/constant"
+	"github.com/tingly-dev/tingly-box/internal/protocol"
+	"github.com/tingly-dev/tingly-box/internal/typ"
+)
+
+// ProviderRecord is the GORM model for persisting a complete provider
+// This includes both configuration and credentials as one logical entity
+type ProviderRecord struct {
```
FFengIll, line 24: take care of this, I think this db schema is too bad.

FFengIll: @andreasfoo

andreasfoo: comment the interceptor plz.

PR #382, opened 2026-02-27
Fix path splitting bug and implement skill.md parser

richardxx:
## Summary

This PR fixes two issues:

### 1. Fixed Path Splitting Bug
- **Problem:** `filepath.SplitList(relPath)` was incorrectly used to split path components
- **Solution:** Changed to `strings.Split(relPath, string(filepath.Separator))`
- **Reason:** `filepath.SplitList` is designed for PATH-like environment variables (using `:` on Unix), not for splitting file paths into components

### 2. Implemented Skill.md Parser
- Added `parseSkillDescription()` function to extract descriptions from skill markdown files
- **Primary method:** Extracts text from the first markdown heading (e.g., `# Skill Name`)
- **Fallback method:** Uses first non-empty paragraph if no heading found
- **Smart filtering:** Skips code blocks, lists, HTML tags, and links
- **Length limiting:** Truncates descriptions to 200 characters with "..." suffix

### Integration Points
- `scanDirectoryForSkills()`: Now populates `Description` field during skill scanning
- `GetSkillContent()`: Populates `Description` when loading skill content

### Tests Added
- `TestParseSkillDescription`: Unit tests for the description parser
- `TestScanLocation_WithDescription`: Integration test verifying descriptions are extracted during scanning

All tests pass successfully.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/internal/data/skill_manager.go
+++ b/internal/data/skill_manager.go
@@ -403,6 +403,22 @@ func resolvePath(homeDir, path string) string {
 // It looks for:
 // 1. First heading (e.g., "# Skill Name") - returns the heading text
 // 2. First non-empty paragraph - returns the paragraph text (truncated if too long)
+//
+// Examples:
+//   Input: "# Commit Skill\n\nThis is a commit skill."
+//   Output: "Commit Skill"
+//
+//   Input: "## Code Review\n\nThis skill helps with reviews."
+//   Output: "Code Review"
+//
+//   Input: "No heading here, just a description."
+//   Output: "No heading here, just a description."
+//
+//   Input: "```go\nfunc main() {}\n```\n\nActual description here."
+//   Output: "Actual description here."
+//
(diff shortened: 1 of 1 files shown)


internal/data/skill_manager.go
```diff
@@ -389,6 +399,80 @@ func resolvePath(homeDir, path string) string {
 	return filepath.Join(homeDir, path)
 }
 
+// parseSkillDescription extracts a description from skill markdown content.
+// It looks for:
+// 1. First heading (e.g., "# Skill Name") - returns the heading text
+// 2. First non-empty paragraph - returns the paragraph text (truncated if too long)
+func parseSkillDescription(content string) string {
```
FFengIll, line 422: please add example in comment and let us understand the logic clearly.

PR #383, opened 2026-02-27
feat/agentboot: update Claude Code boot adapter to support user approval

FFengIll:
This PR improved claude code stream-json interaction to support user approval.
Furthermore, we update some logic in agentboot
- new permission handler to adapter different agents for future.
- new claude code message formatter.
- new agent execution config for timeout.

### DIFF
--- a/agentboot/claude/examples/server/server.go
+++ b/agentboot/claude/examples/server/server.go
@@ -134,7 +134,6 @@ func (s *Server) ProcessQuery(ctx context.Context, userPrompt string, continueCo
 	var response strings.Builder
 	var assistantText strings.Builder
 
-messageLoop:
 	for {
 		msg, ok := query.Next()
 		if !ok {
@@ -185,8 +184,7 @@ messageLoop:
 				response.WriteString(msg.Result)
 			}
 			// Result message indicates the turn is complete
-			// Use labeled break to exit the for loop
-			break messageLoop
+			goto done
 		case "control_request", "control_response", "control_cancel_request":
 			// These are handled internally by the Query
 			log.Printf("[Control] %s message handled internally", msg.Type)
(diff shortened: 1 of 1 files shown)


agentboot/claude/examples/server/server.go
```diff
@@ -0,0 +1,387 @@
...
+			// Tool result is handled internally by Claude CLI
+		case "result":
+			log.Printf("[Result] Result: %s", msg.Result)
+			if msg.Result != "" {
+				// Always append result - it contains the final response
+				response.WriteString(msg.Result)
+			}
+			// Result message indicates the turn is complete
+			// Use labeled break to exit the for loop
+			break messageLoop
```
andreasfoo: what about just goto line 198

agentboot/claude/control.go
```diff
@@ -0,0 +1,332 @@
...
+	if err := m.writeRequest(stdin, req); err != nil {
+		return ControlResponse{}, fmt.Errorf("write control request: %w", err)
+	}
+
+	// Wait for response
+	select {
+	case resp := <-respChan:
+		return resp, nil
+	case <-timeoutCtx.Done():
+		return ControlResponse{}, fmt.Errorf("control request timeout after %v", m.requestTimeout)
```
andreasfoo: maybe we should tell the user timeout is xx secs or min?

agentboot/claude/control.go
```diff
@@ -0,0 +1,332 @@
...
+// IsClosed returns true if the manager is closed
+func (m *ControlManager) IsClosed() bool {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	return m.closed
+}
+
+// generateRequestID generates a unique request ID
+func generateRequestID() string {
+	return fmt.Sprintf("req_%d", time.Now().UnixNano())
```
andreasfoo: hard to collide but strange to use timestamp as id...

agentboot/claude/examples/server/server.go
FFengIll: change to goto.

agentboot/claude/control.go
FFengIll: change to UUID.

PR #420, opened 2026-03-05
feat: add new flag to control thinking effort.

FFengIll:


frontend/src/pages/scenario/UseOpenCodePage.tsx
```diff
@@ -204,7 +204,7 @@ const UseOpenCodePage: React.FC = () => {
                                 variant="contained"
                                 size="small"
                             >
-                                Config OpenCode
+                                Quick Config
```
seviezhou, line 207: why change this name?

frontend/src/i18n/locales/en.ts
```diff
@@ -374,7 +374,7 @@ export default {
     "separateConfig": "Separate Configuration",
     "switchToSeparate": "Switch to Separate",
     "switchToUnified": "Switch to Unified",
-    "configButton": "Configuration Guide",
+    "configButton": "Quick Config",
```
seviezhou, line 377: should we use unified names for configButton

frontend/src/pages/scenario/UseOpenCodePage.tsx
FFengIll: unified them in each page.

PR #422, opened 2026-03-06
feat: add scenario for Codex

seviezhou:
## Background
This PR adds end-to-end support for `Scenario Codex`.  
Some provider/model combinations do not support the `/responses` endpoint, so Codex requests must be routed through capability-aware fallback paths while preserving Responses-compatible behavior.

## What Changed
- Added Codex scenario handling in `internal/server/openai_responses.go`:
  - Detect provider/model endpoint capability.
  - Use `/responses` when supported.
  - Fallback to Chat/Anthropic forwarding when Responses is unsupported.
- Integrated new request/stream converters:
  - `openai_responses_to_chat`
  - `openai_responses_to_anthropic`
  - `openai_chat_to_responses`
- Fixed Chat -> Responses streaming event compatibility:
  - Added required fields (notably `sequence_number`).
  - Added missing lifecycle events:
    - `response.output_item.added/done`
    - `response.output_text.delta/done`
    - `response.function_call_arguments.delta/done`
  - Fixed `response.completed.usage` shape, including `total_tokens` and token detail objects.
  - Added completion fallback for providers that end streams without explicit finish events (`response.completed` + `[DONE]`).

## Issues Addressed
- `responses_not_supported` failures for Codex on non-Responses providers.
- Stream parse error: `missing field total_tokens`.
- “No visible reply” in Codex despite stream handler finishing (caused by incomplete/invalid Responses event schema).

iwEdAqNwbmcDAQTRCzoF0QaYBrAsxknIkcbpywmHTWRR5iQAB9IcRCLUCAAJomltCgAL0gAF9MY

### DIFF
--- a/frontend/src/pages/scenario/UseCodexPage.tsx
+++ b/frontend/src/pages/scenario/UseCodexPage.tsx
@@ -8,7 +8,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import EmptyStateGuide from '@/components/EmptyStateGuide';
 import PageLayout from '@/components/PageLayout';
-import TemplatePage from '@/components/TemplatePage.tsx';
+import TemplatePage from './components/TemplatePage.tsx';
 import { useFunctionPanelData } from '@/hooks/useFunctionPanelData';
 import { useHeaderHeight } from '@/hooks/useHeaderHeight';
 import { api, getBaseUrl } from '@/services/api';


frontend/src/pages/scenario/UseCodexPage.tsx
```diff
@@ -0,0 +1,147 @@
+import CardGrid from "@/components/CardGrid.tsx";
```
FFengIll, line 1: please up to date and check if codex page missing some plugins.

PR #478, opened 2026-03-14
cli: add start hook to register enterprise routes

andreasfoo:
## Problem
`tbe start` overrides the default start path and bypasses `startServer`, so the `multiLogger` is never injected. This causes a nil pointer panic when `NewMultiModeMemoryLogMiddleware` accesses `multiLogger`.

## Change
- Add `StartCommandWithHook` / `startServerWithHook` to allow a setup hook after the server manager is created and before the server starts.
- Keep `StartCommand` behavior unchanged by delegating to the hook-enabled variant with a nil hook.

## Why
This keeps enterprise route registration extensible without skipping the core startup flow that wires logging.

## Testing
Manual: start `tbe` and confirm no `MultiLogger` nil panic.

### DIFF
--- a/internal/command/server.go
+++ b/internal/command/server.go
@@ -93,11 +93,11 @@ func doStopServer(appManager *AppManager) error {
 
 // startServer handles the server starting logic
 func startServer(appManager *AppManager, opts options.StartServerOptions) error {
-	return startServerWithHook(appManager, opts, nil)
+	return startServerWithHook(appManager, opts)
 }
 
-// startServerWithHook handles the server starting logic with an optional setup hook.
-func startServerWithHook(appManager *AppManager, opts options.StartServerOptions, hook func(*ServerManager) error) error {
+// startServerWithHook handles the server starting logic with optional setup hooks.
+func startServerWithHook(appManager *AppManager, opts options.StartServerOptions, hooks ...func(*ServerManager) error) error {
 	appConfig := appManager.AppConfig()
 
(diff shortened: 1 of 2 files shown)


internal/server/config/config.go
```diff
@@ -373,15 +373,29 @@ func (c *Config) Save() error {
 	if c.ConfigFile == "" {
 		return fmt.Errorf("ConfigFile is empty")
 	}
-	data, err := json.MarshalIndent(c, "", "    ")
+	data, err := json.Marshal(c)
 	if err != nil {
 		return err
 	}
-	err = os.WriteFile(c.ConfigFile, data, 0644)
+	var next map[string]interface{}
```
FFengIll, line 380: why do marshal / unmarshal twice here?

internal/command/server.go
```diff
@@ -93,6 +93,11 @@ func doStopServer(appManager *AppManager) error {
 
 // startServer handles the server starting logic
 func startServer(appManager *AppManager, opts options.StartServerOptions) error {
+	return startServerWithHook(appManager, opts, nil)
+}
+
+// startServerWithHook handles the server starting logic with an optional setup hook.
+func startServerWithHook(appManager *AppManager, opts options.StartServerOptions, hook func(*ServerManager) error) error {
```
FFengIll: suggest to change to hooks via vargs.

FFengIll (dismissed):

PR #501, opened 2026-03-18
bugfix/codex: improve codex support with protocol process enhance and http process bugfix.

FFengIll:
## Summary

Implements ChatGPT/Codex OAuth support using a protocol transformer chain architecture, enabling seamless integration between the ChatGPT backend API and OpenAI/Anthropic formats.

## Major: Codex OAuth Client
- Introduced `codexRoundTripper` to transform Codex backend API responses into the OpenAI Responses API format.
- Features include URL path rewriting, header transformation, and request body filtering (removes unsupported parameters).
- Performs on-the-fly response format transformation for both streaming and non-streaming modes.

## Major: Protocol Transformer Chain
- Refactored OAuth provider handling to use a `RoundTripper` wrapper chain: base → proxy → request modifier → response transformer.
- Unified `CreateHTTPClientForProvider` with provider-specific transport configurations.
- Antigravity and Codex now utilize specialized `RoundTripper` implementations.

## Major: Stream & Server Handler Unification
- Added `HandleOpenAIResponsesStreamToAnthropic` for transforming Codex backend streams to the Anthropic format.
- The Codex backend employs an "assembly" pattern: streams internally, then assembles into a non-streaming response.
- Unified core streaming logic with `handlerResponsesToAnthropicStream`, shared between v1 and beta endpoints.
- Removed legacy ChatGPT backend handlers (`codex_responses.go`, `chatgpt_backend.go`).

## Minor: Cleanup
- Removed the hook-based `GetOAuthHook` function.
- Added context cancellation checks before stream iteration.

## Test Plan
- Verify the Codex OAuth authentication flow.
- Test streaming and non-streaming responses in both v1 and beta formats.
- Validate proxy configuration with the new transport chain.

### DIFF
--- a/internal/client/openai.go
+++ b/internal/client/openai.go
@@ -157,7 +157,7 @@ func (c *OpenAIClient) ListModels(ctx context.Context) ([]string, error) {
 		}
 	}
 	// Also handle legacy ChatGPT backend API providers
-	if c.provider.APIBase == protocol.ChatGPTBackendAPIBase {
+	if c.provider.APIBase == protocol.CodexAPIBase {
 		return nil, &ErrModelsEndpointNotSupported{
 			Provider: c.provider.Name,
 			Reason:   "ChatGPT backend API does not support /models endpoint",
--- a/internal/client/codex_e2e_test.go
+++ b/internal/client/codex_e2e_test.go
@@ -43,7 +43,7 @@ func TestE2E_CodexRoundTripper(t *testing.T) {
 	provider := &typ.Provider{
 		ProxyURL: "socks5://127.0.0.1:7890",
 		Name:     "codex-e2e-test",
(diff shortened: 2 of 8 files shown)


internal/client/openai.go
```diff
@@ -157,7 +157,7 @@ func (c *OpenAIClient) ListModels(ctx context.Context) ([]string, error) {
 		}
 	}
 	// Also handle legacy ChatGPT backend API providers
-	if c.provider.APIBase == protocol.ChatGPTBackendAPIBase {
+	if c.provider.APIBase == protocol.CodexAPIBase {
```
andreasfoo, line 160: why this？ i know oauth do not support list models but why the old code says chatgptbackendapibase do not support? -- just renaming?

FFengIll: just renaming.

PR #526, opened 2026-03-21
Fix Codex routing and fallback behavior for provider/model endpoint mismatches

seviezhou:
## Problem
Codex requests were failing for some provider/model combinations even when the same provider passed connection checks.

The failure mode was confusing:

- `test connection` succeeded
- the provider could serve the model through `chat/completions`
- but Codex still failed with errors such as:
  - `Stream disconnected before completion: Unsupported model`
  - `context deadline exceeded`

After tracing the request path, there were actually two separate issues.

### 1. Connection checks did not validate the Codex path
The existing connection test only verified:

- `GET /models`
- `POST /chat/completions`

Codex, however, primarily uses the Responses API path:

- `POST /responses`
- often with `stream=true`

This created a false sense of compatibility:
a provider could pass connection checks and still fail for the actual Codex request path.

### 2. Endpoint capability decisions were too optimistic
Codex routing depended on the stored preferred endpoint for a given `provider + model`.

In practice, some models behaved like this:

- `chat/completions` works
- non-streaming `responses` may work
- streaming `responses` fails immediately with `response.failed`
- upstream error is `Unsupported model`

```
INFO[0135] Evaluating Codex responses routing            actual_model=glm-4.7 api_base="https://dashscope.aliyuncs.com/compatible-mode/v1" preferred_endpoint=responses provider=qwen-plus request_model=tingly-codex scenario=codex stream=true
```

This meant a provider/model pair could be marked as preferring `responses`, but the real streaming Codex request still failed at runtime.

### 3. Fallback streaming requests could immediately time out
When Codex fell back from `responses` to `chat`, some providers had `timeout=0`.

```
ERRO[1556] Chat to Responses stream error: context deadline exceeded  cache_tokens=0 input_tokens=0 output_tokens=0 response_id=resp_1774110803 response_model=tingly-codex text_len=0 tool_calls=0
```

That value was passed through to `context.WithTimeout(...)`, which caused the fallback request context to expire immediately and produced:

- `Chat to Responses stream error: context deadline exceeded`

## Root Cause
The main issue was not a single bad model mapping. It was the combination of:

- connection checks testing `chat` but not real Codex `responses`
- routing choosing `responses` based on incomplete capability information
- no automatic recovery when upstream returned `response.failed: Unsupported model`
- zero timeout values being treated as literal request deadlines instead of "use default"

## Solution
This PR improves Codex resilience in three ways.

### 1. Add detailed request and stream diagnostics
Added structured logs for:

- resolved request model vs actual upstream model
- selected provider and API base
- preferred endpoint decision
- whether Codex used native `responses` or `chat` fallback
- full `response.failed` payload from upstream
- stream termination state:
  - event count
  - last event
  - whether `response.completed` was ever received

This makes it possible to distinguish:

- upstream transport errors
- application-level `response.failed`
- early stream termination without SDK error
- endpoint selection mistakes

### 2. Add automatic fallback from `responses` to `chat`
For Codex streaming requests:

- try native streaming `responses` first
- inspect the first upstream event
- if the first event is `response.failed`
- and the upstream error message contains `Unsupported model`
- immediately close the `responses` stream
- transparently retry through the existing `chat -> responses` fallback path

This preserves Codex behavior for providers that support the model via chat but not via streaming responses.

### 3. Fix zero-timeout handling
Updated `ForwardContext` timeout initialization so that:

- provider-specific timeout is used when configured
- `timeout <= 0` now falls back to `constant.DefaultRequestTimeout`

This prevents fallback streaming requests from failing immediately with `context deadline exceeded`.

## Result
After this change:

- Codex no longer hard-fails when a provider rejects a model on streaming `responses` but supports it through `chat`
- zero-timeout providers no longer break fallback streaming
- logs clearly show whether the failure is caused by:
  - endpoint mismatch
  - upstream `response.failed`
  - missing `response.completed`
  - timeout configuration

## Files Changed
- `internal/server/openai_responses.go`
- `internal/server/protocol_forward.go`
- `internal/server/protocol_context.go`
- `internal/protocol/stream/openai_passthrough.go`

## Validation
- `GOCACHE=/tmp/go-build go test ./internal/server -run TestNonExistent -count=1`

## Notes
This PR does not yet redesign the capability probe model to distinguish between:

- non-streaming `responses`
- streaming `responses`

It instead adds runtime recovery and better diagnostics, which is sufficient to make Codex usable for mixed-compatibility providers.

### DIFF
--- a/internal/server/openai_responses.go
+++ b/internal/server/openai_responses.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/anthropics/anthropic-sdk-go"
 	"github.com/gin-gonic/gin"
@@ -135,29 +134,8 @@ func (s *Server) ResponsesCreate(c *gin.Context) {
 		return
 	}
 
-	logrus.WithFields(logrus.Fields{
-		"scenario":      scenarioType,
-		"request_model": string(req.Model),
-		"actual_model":  actualModel,
-		"params_model":  string(params.Model),
-		"provider":      provider.Name,
-		"provider_uuid": provider.UUID,
-		"api_base":      provider.APIBase,
-		"api_style":     provider.APIStyle,
-		"stream":        req.Stream,
-	}).Info("Resolved Responses request")
-
 	if scenarioType == typ.ScenarioCodex && provider.APIBase != protocol.CodexAPIBase {
 		preferredEndpoint := NewAdaptiveProbe(s).GetPreferredEndpoint(provider, actualModel)
-		logrus.WithFields(logrus.Fields{
-			"scenario":           scenarioType,
-			"request_model":      string(req.Model),
-			"actual_model":       actualModel,
-			"provider":           provider.Name,
-			"api_base":           provider.APIBase,
-			"stream":             req.Stream,
-			"preferred_endpoint": preferredEndpoint,
-		}).Info("Evaluating Codex responses routing")
 		if preferredEndpoint != "responses" {
 			s.handleCodexResponsesFallback(c, provider, params, req.Model, actualModel, maxAllowed, req.Stream)
 			return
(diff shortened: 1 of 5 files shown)


internal/server/openai_responses.go
FFengIll, line 1: 根据用户反馈，codex发出的是 openai responses 请求，而得到了服务端报错 provider prefer chat。
这说明，我们没有正确的转换 request 从 responses 到 chat，并在服务逻辑中卡住了。
因此，还需要检查prefer的检查和处理逻辑。

PR #542, opened 2026-03-24
Use injected auth middleware for web UI routes

andreasfoo:
## Summary
- route web UI user-auth registration through the injectable auth accessor
- ensure TBE-injected RBAC auth applies to OAuth web UI routes
- ensure load-balancer management routes also use injected user auth

## Testing
- go test ./cli/tbe -run ^$


FFengIll (approved):
Glad to see the fix.

PR #565, opened 2026-03-26
feat: add scenario registration registry

andreasfoo:
## Summary
- add a minimal scenario registration registry in `internal/typ`
- allow rule create/update to bind only to registered, bindable scenarios
- keep direct path usage gated separately so enterprise-only scenarios can be registered without changing existing routing semantics

## What This Does Not Change
- no fallback lookup
- no built-in migration
- no scenario uniqueness changes
- no `general` builtin behavior
- no mixed scenario listing behavior

## Verification
- `go test ./internal/typ`
- `go test ./internal/server -run ^$`

## Notes
- `internal/server/module/rule` has pre-existing test-file drift against the current `NewHandler` signature and was not expanded as part of this PR.

### DIFF
--- a/internal/typ/scenario_registry.go
+++ b/internal/typ/scenario_registry.go
@@ -1,0 +1,156 @@
+package typ
+
+import (
+	"fmt"
+	"slices"
+	"sync"
+)
+
+type ScenarioTransport string
+
+const (
+	TransportOpenAI    ScenarioTransport = "openai"
+	TransportAnthropic ScenarioTransport = "anthropic"
+)
+
+type ScenarioDescriptor struct {
+	// ID is the stable scenario identifier stored on rules and scenario configs.
+	ID RuleScenario `json:"id" yaml:"id"`
+	// SupportedTransport declares which protocol surfaces may resolve rules bound to this scenario.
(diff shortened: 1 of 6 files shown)


internal/typ/scenario_registry.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+type ScenarioTransport string
+
+const (
+	TransportOpenAI    ScenarioTransport = "openai"
+	TransportAnthropic ScenarioTransport = "anthropic"
+)
+
+type ScenarioDescriptor struct {
```
FFengIll, line 16: please add comment for all fields to explain,

internal/typ/type.go
```diff
@@ -62,6 +62,21 @@ const (
 	ScenarioGlobal     RuleScenario = "_global" // Global flags that apply to all scenarios
 )
 
+func BuiltinRuleScenarios() []RuleScenario {
```
FFengIll: BuiltinScenarios ?

internal/typ/scenario_registry.go
```diff
@@ -0,0 +1,156 @@
...
```
FFengIll, line 16: We may remove this in future after disable tb scenario validation.

PR #616, opened 2026-04-02
refactor(server): migrate openai chat to use transform chain dispatch

FFengIll:
## Summary
OpenAI Chat Completions was using a separate code path instead of the unified transform chain dispatch. This created inconsistency and made it harder to support cross-protocol conversions like Chat→Responses.

part of #500 

### Major
- OpenAI Chat now uses the same dispatch chain as other protocols
- Added Responses→Chat conversion for providers using Responses API with Chat clients
- Chat→Responses→Chat conversion path is now possible (an error converson before)

### Minor
- Added `stream.HandleResponsesToOpenAIChatStream()` for streaming Responses→Chat conversion
- Added `nonstream.OpenAIResponsesToChat()` for non-streaming conversion
- Reduced `openai.go` by ~120 lines through dispatch consolidation
- Added `dispatchOpenAIChatFromResponses()` handler in `protocol_dispatch.go`

### DIFF
--- a/internal/protocol/nonstream/openai_reponses_to_chat.go
+++ b/internal/protocol/nonstream/openai_reponses_to_chat.go
@@ -1,0 +1,47 @@
+package nonstream
+
+import "github.com/openai/openai-go/v3/responses"
+
+// OpenAIResponsesToChat converts a Responses API response to Chat Completions format.
+// This is used when the client expects Chat format but the provider uses Responses API.
+func OpenAIResponsesToChat(resp *responses.Response, responseModel string) map[string]any {
+	// Extract text content from output
+	// Response output can have various types: "message", "function_call", etc.
+	content := ""
+	for _, output := range resp.Output {
+		if output.Type == "message" {
+			// For message type, extract text from content array
+			for _, contentItem := range output.Content {
+				// ContentItemUnion has a Type field to distinguish variants
+				if contentItem.Type == "output_text" || contentItem.Type == "text" {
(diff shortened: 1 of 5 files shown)


internal/protocol/nonstream/openai_reponses_to_chat.go
```diff
@@ -0,0 +1,47 @@
...
+		"id":      resp.ID,
+		"object":  "chat.completion",
+		"created": int64(resp.CreatedAt),
+		"model":   responseModel,
+		"choices": choices,
+		"usage": map[string]any{
+			"prompt_tokens":     resp.Usage.InputTokens,
+			"completion_tokens": resp.Usage.OutputTokens,
+			"total_tokens":      resp.Usage.InputTokens + resp.Usage.OutputTokens,
+		},
```
andreasfoo, line 45: No tool call? And why the stop reason is always 'stop' ?

internal/protocol/nonstream/openai_reponses_to_chat.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Build Chat Completion response structure
+	choices := []map[string]any{
+		{
+			"index": 0,
+			"message": map[string]any{
+				"role":    "assistant",
+				"content": content,
+			},
+			"finish_reason": "stop",
```
FFengIll, line 31: @seviezhou pls check finish reasson

PR #656, opened 2026-04-09
Refactor guardrails runtime, pipelines, and Anthropic stream handling

seviezhou:
## Summary

This PR restructures the guardrails implementation around a clearer split between runtime wiring in `internal/server` and shared guardrails logic in `internal/guardrails`.

The main changes are:

- turn `guardrails.Guardrails` into the unified runtime surface
- move shared guardrails logic into `core`, `evaluate`, `adapter`, `mutate`, and `pipeline`
- remove the old `internal/server/guardrails` compatibility layer
- simplify request, non-stream, and stream guardrails flows around a single `guardrailscore.Input`
- reduce stream-specific state spread by moving guardrails stream state onto `protocol.HandleContext`

### Runtime and package structure

- `internal/guardrails` is now organized around:
  - `core`: shared models and semantic types
  - `evaluate`: evaluation engine and policy implementations
  - `adapter`: protocol-specific normalization/parsing
  - `mutate`: protocol mutation helpers
  - `pipeline`: shared request/response orchestration
  - `utils`: history and credential store utilities
- `guardrails.Guardrails` now represents the runtime, rather than just an evaluate interface.
- most root-package wrapper/alias files were removed; the root package is now much closer to a façade/runtime package.

### Input model unification

- `guardrailscore.Input` is now the single guardrails processing object.
- request/response `View` types were removed:
  - `RequestView`
  - `ToolResultRequestView`
  - `ResponseView`
  - `WithRequestView`
  - `WithResponseView`
  - `SessionMeta`
  - `ToRequestInput`
  - `ToResponseInput`
- adapters now refresh/fill `Input` directly.
- request-side tool result hints live directly on `Input`:
  - `HasToolResult`
  - `ToolResultBlockCount`
  - `ToolResultPartCount`

### Request pipelines

- Anthropic v1 and beta request guardrails now run through shared pipeline helpers.
- request processing order was changed to:
  1. refresh input from raw request
  2. apply credential mask
  3. refresh input
  4. evaluate tool_result as a post-mask fallback
- request pipeline result structs that only served as intermediate wrappers were removed.

### Non-stream response guardrails

- non-stream response handling was pushed into shared pipeline code.
- server-side non-stream wrappers now only do server wiring.
- `applyGuardrailsToAnthropic*NonStreamResponse(...)` now owns:
  - `ensureGuardrailsCredentialMaskState(...)`
  - input wiring
  - alias restore on non-blocked responses

### Stream guardrails

- stream guardrails now only perform early `tool_use` interception.
- the final content-wide `onStreamComplete` evaluation was removed.
- tool_use buffering and pending block state no longer live in `gin.Context`; they now live on `protocol.HandleContext.Guardrails.Stream`.
- `HandleContext` guardrails fields were grouped into:
  - `Guardrails.Enabled`
  - `Guardrails.CredentialMask`
  - `Guardrails.Stream`
- `rewriteAnthropicGuardrailsEvent(...)` was simplified:
  - only intercepts tool_use-related events
  - no text alias restore during stream rewrite
  - no fallback synthetic block injection
  - fewer wrapper/helper layers

### Command parsing and adapters

- protocol-side command construction was moved toward `adapter`.
- adapter now provides `BuildCommand(...)` and `BuildCommandFromRawArguments(...)`.
- stream hooks and Anthropic response adapters no longer manually assemble commands in server code.

### Credential mask state

- `CredentialMaskState` is back to the direct `gin.Context` key pattern using `CredentialMaskStateContextKey`.
- `TransformContext` no longer carries guardrails credential state.
- `Input.State.CredentialMask` remains the explicit pipeline input, with `gin.Context` as a fallback source.

## Cleanup

- removed the old `internal/server/guardrails` package
- removed `internal/server/guardrails_nonstream.go`
- removed `internal/server/guardrails_history.go`
- removed many temporary wrappers and compatibility aliases
- updated `docs/guardrails-architecture.md` to reflect the new structure

## Result

The net effect is:

- less duplicated adapter/evaluate logic
- fewer server-side compatibility wrappers
- clearer ownership of request-scoped vs runtime-scoped guardrails state
- a simpler Anthropic stream interception model
- a more coherent package layout for future guardrails work


seviezhou:
Merge #684 into this pr.

richardxx:
## PR #656 Code Review — Guardrails Refactor

Comprehensive review found **1 compile-blocking issue, 4 high-severity bugs, and several medium/low issues**.

### Compile-blocking
- **`*bool` → `bool` type mismatch in `registry_test.go`** — `&enabled` is `*bool` but `PolicyGroup.Enabled` is now plain `bool`. Won't compile.

### High Severity

**1. Behavioral regression: `Enabled *bool → bool` silently disables all policies by default** (`core/policy.go:19`)
The old `*bool` allowed distinguishing "field omitted" (nil → default enabled) from "explicitly false". Now `bool` defaults to `false`. The old `resolvePolicyEnabled` defaulted to `true` when nil; the new code returns `false` for zero-value. All policies that previously defaulted to enabled now silently become disabled when YAML omits the field. **Revert to `*bool`.**

**2. Missing guardrails on non-Anthropic dispatch paths** (`protocol_dispatch.go`)
Guardrails hooks are only attached in `dispatchChainResultToAnthropicV1` and `dispatchChainResultToAnthropicV1Beta`. These paths have **no guardrails evaluation**:
- `dispatchChainResultToGoogle` — Google backend
- `dispatchChainResultToResponses` — OpenAI Responses API backend (Codex)
- `dispatchChainResultToOpenAIChat` — OpenAI Chat backend

Tool_use interception, credential masking, and content blocking are silently skipped for any Anthropic/Claude Code request routed through these backends.

**3. Only FIRST tool_use block is extracted** (`adapter/anthropic_v1.go:57`, `anthropic_beta.go:57`)
`commandFromAnthropicV1Blocks` returns on the first `tool_use` match. When a model response contains multiple tool calls (common in agentic workflows), only the first is evaluated by guardrails. The rest bypass policy checking entirely. Same issue in the `StreamAccumulator.Content()` — flat fields overwrite per-index tracking.

**4. V1 thinking block extraction uses wrong field** (`adapter/anthropic_v1.go:49`)
```go
if (block.Type == "text" || block.Type == "thinking") && strings.TrimSpace(block.Text) != "" {
```
For `block.Type == "thinking"`, the content is in `block.Thinking`, not `block.Text`. The beta version correctly uses `block.Thinking`. Thinking content is silently dropped for V1 responses.

### Medium Severity

**5. `ContentPolicy` blocks ALL input for credential-only policies** (`evaluate/content_policy.go:79`)
When a policy has `CredentialRefs` but no `Patterns`, the evaluation returns the verdict immediately without checking if credentials are actually present in the text.

**6. `mergePolicyScope` is now identity — group default scope/verdict inheritance lost** (`evaluate/registry.go:70`)
The old version merged `group.DefaultScope` into the policy scope. Now it just returns the input unchanged. `PolicyGroup.DefaultScope` and `DefaultVerdict` were removed from the struct.

**7. `ReplaceToolResultContentV1` replaces ALL tool_result blocks** (`adapter/request_helpers.go:80`)
Iterates over every message and replaces every `tool_result`. If a conversation has 5 tool results and only the latest triggers a block, all 5 get replaced. Should be scoped to the offending message/block only.

**8. Data race on `s.guardrailsRuntime`** (`guardrails_runtime.go`)
`setGuardrailsRuntime` reads and writes `s.guardrailsRuntime` without any lock. Concurrent config reloads can race on the pointer and `History`/`CredentialCache` fields.

### Low Severity

**9. Credential masking misses `tool_use.Name`** (`mutate/anthropic_request_mask.go`)
Masks text, tool result content, and tool use arguments, but not `block.OfToolUse.Name`.

**10. `GuardrailsStreamState` maps not thread-safe** (`protocol/context.go:50`)
4 map fields with no mutex. Currently safe due to single-threaded streaming, but fragile.

**11. `onStreamComplete` evaluation removed** (design decision)
Only early `tool_use` interception remains for streaming. Malicious content in text blocks is no longer caught during streaming. Non-stream responses still have full evaluation.

### Recommended Fix Priority
1. Revert `Enabled` to `*bool` (fixes compile error + behavioral regression)
2. Add guardrails hooks to Google/OpenAI dispatch paths
3. Fix tool_use extraction to evaluate ALL blocks
4. Fix V1 thinking block field access
5. Add lock around `s.guardrailsRuntime`
6. Scope `ReplaceToolResultContentV1` to the offending block only

richardxx (commented):
## Guardrails Refactoring PR Review

Automated review of the guardrails refactoring. Findings organized by severity.

### Summary
- **HIGH**: 7 issues (compilation errors, broken logic, missing guardrails, data races)
- **MEDIUM**: 5 issues (incomplete extraction, missing masking, thread safety)

See inline comments for details.

internal/guardrails/core/policy.go
```diff
@@ -1,4 +1,4 @@
-package guardrails
+package core
 
 // PolicyKind is the user-facing policy classification.
 type PolicyKind string
```
richardxx, line 4: **[HIGH] `Enabled` field changed from `*bool` to `bool` breaks default-enabled behavior.**

Previously, `nil` meant "use the parent default" and `true`/`false` were explicit overrides. Now a zero-value `false` is indistinguishable from "explicitly disabled." Any policy YAML that omits `enabled:` (relying on default-true) will now be treated as disabled.

**Fix:** Revert to `*bool` or add explicit `DefaultEnabled bool` logic in YAML unmarshaling.

internal/guardrails/evaluate/registry_test.go
```diff
@@ -1,32 +1,36 @@
...
+
+	guardrailscore "github.com/tingly-dev/tingly-box/internal/guardrails/core"
+)
 
 func TestBuildEvaluatorsCreatesResourceAccessPolicyEvaluator(t *testing.T) {
 	enabled := true
-	evaluators, err := BuildEvaluators(Config{
-		Groups: []PolicyGroup{
+	evaluators, err := BuildEvaluators(guardrailscore.Config{
+		Groups: []guardrailscore.PolicyGroup{
```
richardxx, line 12: **[HIGH] Compilation error: `&enabled` on a `bool` literal.**

`Enabled` was changed from `*bool` to `bool` in `policy.go`, but this test still takes its address with `&enabled`. Go does not allow taking the address of a literal `true`. This file will not compile.

**Fix:** Change to just `enabled: true` (value, not pointer) after confirming the `Enabled` field type change is intentional.

internal/guardrails/adapter/anthropic_v1.go
```diff
@@ -0,0 +1,78 @@
...
+		Messages: input.Content.Messages,
+	}
+	if resp != nil {
+		input.Content.Text = responseTextFromAnthropicV1Blocks(resp.Content)
+		input.Content.Command = commandFromAnthropicV1Blocks(resp.Content)
+	}
+	return input
+}
+
+func responseTextFromAnthropicV1Blocks(blocks []anthropic.ContentBlockUnion) string {
```
richardxx, line 49: **[HIGH] `responseTextFromAnthropicV1Blocks` reads `block.Text` for thinking blocks instead of `block.Thinking`.**

For Anthropic V1 API, thinking blocks use the `Thinking` field, not `Text`. This function will return empty strings for all thinking content, which means guardrails content analysis will see nothing to evaluate for thinking blocks.

**Fix:** Use `block.Thinking` when `block.Type` is `"thinking"`.

internal/guardrails/adapter/anthropic_v1.go
```diff
@@ -0,0 +1,78 @@
...
+	parts := make([]string, 0, len(blocks))
+	for _, block := range blocks {
+		if (block.Type == "text" || block.Type == "thinking") && strings.TrimSpace(block.Text) != "" {
+			parts = append(parts, block.Text)
+		}
+	}
+	return strings.Join(parts, "\n")
+}
```
richardxx, line 57: **[MEDIUM] `commandFromAnthropicV1Blocks` only extracts the first `tool_use` block.**

If the model returns multiple tool_use blocks in a single response (e.g., parallel tool calls), only the first one is checked by guardrails. The others pass through unvalidated.

**Fix:** Return all tool_use blocks, or at minimum document that parallel tool calls are not individually guarded.

internal/guardrails/adapter/anthropic_beta.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	if resp != nil {
+		input.Content.Text = responseTextFromAnthropicV1BetaBlocks(resp.Content)
+		input.Content.Command = commandFromAnthropicV1BetaBlocks(resp.Content)
+	}
+	return input
+}
+
+func responseTextFromAnthropicV1BetaBlocks(blocks []anthropic.BetaContentBlockUnion) string {
+	parts := make([]string, 0, len(blocks))
```
richardxx, line 49: **[HIGH] Same thinking block bug as V1 adapter.** `block.Text` is used for thinking blocks instead of `block.Thinking`. In the Beta SDK, thinking blocks may use `Thinking` field. This needs verification against the Anthropic SDK version used.

internal/guardrails/adapter/anthropic_beta.go
```diff
@@ -0,0 +1,73 @@
...
+	for _, block := range blocks {
+		switch block.Type {
+		case "text":
+			if strings.TrimSpace(block.Text) != "" {
+				parts = append(parts, block.Text)
+			}
+		case "thinking":
+			if strings.TrimSpace(block.Thinking) != "" {
```
richardxx, line 57: **[MEDIUM] Same single-tool_use extraction bug as V1 adapter.** Only the first tool_use is returned for guardrails evaluation.

internal/guardrails/adapter/stream.go
```diff
@@ -0,0 +1,326 @@
...
+	ID    string
+	Name  string
+	Args  string
+}
+
+type StreamAccumulator struct {
+	textBuilder  strings.Builder
+	commandName  string
+	commandArgs  strings.Builder
+	commandFound bool
```
richardxx, line 23: **[HIGH] `Content()` method ignores per-index `toolUses` map.**

The stream adapter builds a `toolUses map[int][]toolUseInfo` but `Content()` concatenates `sb.String()` (plain text) without including any tool_use content. Guardrails will never see tool_use content from streaming responses.

**Fix:** Include tool_use names/inputs in the `Content()` return value so guardrails can evaluate them.

internal/guardrails/evaluate/content_policy.go
```diff
@@ -1,14 +1,16 @@
-package guardrails
+package evaluate
```
richardxx, line 1: **[MEDIUM] Credential-only policies (no patterns configured) block ALL input without checking content.**

If a policy has `detectCredentials: true` but no regex patterns, the evaluation returns a match on any input because the credential detection runs on everything. This means enabling credential detection without patterns effectively blocks all traffic.

**Fix:** Ensure credential-only policies have proper scope or add a guard that requires at least one pattern for blocking behavior.

internal/guardrails/evaluate/registry.go
```diff
@@ -1,85 +1,38 @@
...
+package evaluate
 
-import "fmt"
+import (
+	"fmt"
 
-// ResolveConfig validates and normalizes policy-based guardrails configs.
-func ResolveConfig(cfg Config) (Config, error) {
-	cfg = normalizePolicyConfig(cfg)
-	if !usesPolicyConfig(cfg) {
```
richardxx, line 8: **[HIGH] `mergePolicyScope` is now an identity function — group default scope inheritance lost.**

The merge logic that previously allowed a group-level scope to be inherited by child policies has been removed/simplified. Each policy now only uses its own scope, so setting a default scope at the group level has no effect.

**Fix:** Restore scope inheritance or document that scope must be set per-policy.

internal/guardrails/mutate/anthropic_request_mask.go
```diff
@@ -0,0 +1,147 @@
+package mutate
```
richardxx, line 1: **[MEDIUM] Request mask does not mask `tool_use.Name`.**

The masking logic masks tool_result content and text blocks, but tool_use block names (function names) are passed through unmasked. If a tool name itself contains sensitive data (e.g., `lookup_ssn_for_user_123`), it would leak through guardrails masking.

**Fix:** Add masking for `tool_use.Name` field.

internal/server/protocol_dispatch.go
```diff
@@ -9,12 +9,12 @@ import (
 	"github.com/openai/openai-go/v3"
 	"github.com/openai/openai-go/v3/responses"
 	"github.com/sirupsen/logrus"
+	guardrailsadapter "github.com/tingly-dev/tingly-box/internal/guardrails/adapter"
```
richardxx, line 16: **[HIGH] Missing guardrails on Google/OpenAI/Responses dispatch paths.**

The Anthropic dispatch path calls `guardrailsRuntime` for input/output validation, but the Google, OpenAI, and Responses dispatch paths bypass guardrails entirely. Users sending requests through these endpoints will not have their guardrails policies applied.

**Fix:** Add guardrails middleware/checks to all protocol dispatch paths.

internal/server/guardrails_runtime.go
```diff
@@ -1,56 +1,68 @@
 package server
 
 import (
+	"github.com/anthropics/anthropic-sdk-go"
```
richardxx, line 4: **[HIGH] Data race on `s.guardrailsRuntime` pointer during concurrent reloads.**

The runtime is stored as a pointer on the server struct and swapped during reload. Multiple concurrent requests could read the pointer while it's being swapped, causing a data race. This is not protected by a mutex or atomic.

**Fix:** Use `sync.RWMutex` or `atomic.Pointer` to protect the runtime pointer.

internal/protocol/context.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
+	guardrailscore "github.com/tingly-dev/tingly-box/internal/guardrails/core"
```
richardxx, line 11: **[MEDIUM] `GuardrailsStreamState` maps not thread-safe.**

The stream state holds maps that are read by the streaming handler goroutine and written by the SSE parsing goroutine concurrently. Without synchronization, this is a data race.

**Fix:** Add `sync.RWMutex` protection or use `sync.Map` for the state maps.

internal/guardrails/adapter/anthropic_v1.go
seviezhou: will handle it later

internal/server/protocol_dispatch.go
seviezhou: do it later

internal/guardrails/mutate/anthropic_request_mask.go
seviezhou: we currently do not consider credentials inside tool name

internal/guardrails/adapter/stream.go
seviezhou: this function is not used, delete it.

internal/guardrails/evaluate/registry.go
seviezhou: group scope is not supported anymore. Old config that has scope key will be ignored.

FFengIll (approved):
Though this refactoring is difficult to review, I believe it's acceptable since our guardrails aren't released publicly.

PR #668, opened 2026-04-10
fix: prevent MIME-type issues by 404ing missing static assets

andreasfoo:
## Summary

When a static asset (e.g., `/assets/file.js`) is not found, the current behavior is to fall through and serve `index.html`. This causes MIME-type mismatches where JavaScript is served as HTML, leading to browser errors like "Uncaught SyntaxError: Unexpected token '<'".

## Changes

- Add early check for asset paths in `NoRoute` handler
- Return 404 immediately when static asset is not found
- Prevents incorrect MIME-type responses

## Test Plan

- [x] Build succeeds
- [x] Missing assets return 404 with correct MIME-type
- [x] Existing routes continue to work correctly


andreasfoo:
Closing to fix incorrect logic. The current implementation blocks all /assets/ paths regardless of whether files exist.

PR #683, opened 2026-04-12
feat(agentboot): refactor and simplify agentboot entry to make any agent can be a executor in future.

FFengIll:
This PR simplify agentboot entry and refactor whole claude code project with code clean.


agentboot/claude/driver.go
```diff
@@ -0,0 +1,246 @@
...
+// IsAvailable checks whether the Claude Code CLI binary is present and usable.
+func (d *Driver) IsAvailable() bool {
+	d.mu.RLock()
+	discovery := d.discovery
+	cliPath := d.cliPath
+	d.mu.RUnlock()
+
+	if cliPath != "" && cliPath != "claude" && cliPath != "anthropic" {
+		_, err := os.Stat(cliPath)
+		return err == nil
```
andreasfoo, line 48: given any existing file not named claude/anthropic would return true?

andreasfoo (approved):

agentboot/claude/driver.go
FFengIll: Yep, if you give a cli as entry, we will just use it, even invalid file.

PR #684, opened 2026-04-12
Guardrails: add router supply-chain defenses for malicious fetches, secrets, and known malicious packages

seviezhou:
This PR adds a first pass of client-side guardrails against the router/tool-call supply-chain attacks described in the paper on malicious LLM API routers.

The core idea is to treat tool-call outputs and request content as an attack surface when traffic may pass through an untrusted router. This PR adds built-in policies that block or constrain several concrete attack classes from the paper, especially malicious URL delivery, package-install tampering, and request-side secret exfiltration.

**What This PR Adds**

1. Built-in policy to block known malicious fetch targets from URLhaus
   - Adds a dedicated builtin policy that blocks shell network requests to known malicious non-IP hosts derived from a local URLhaus dataset.
   - The dataset is reduced to `scheme://host[:port]` prefixes and matched with `prefix` semantics so malicious hosts are blocked without storing full paths.

2. Built-in request-side secret egress protections
   - Adds request-direction content policies for common secret formats such as API keys, cloud credentials, private keys, and dotenv-style dumps.
   - Adds broader request-side blocking for bundled credential material and environment secret exports.

3. Built-in filesystem protections for common secret stores
   - Blocks reads of common secret-bearing files such as AWS credentials, cloud config files, private key files, SSH material, dotenv files, and shell history.
   - This pushes enforcement earlier in the flow so secrets are blocked before they can be forwarded to a model router.

4. Built-in malicious package blocklists sourced from OpenSSF Malicious Packages
   - Splits package-related supply-chain rules into dedicated builtin YAML files by ecosystem.
   - Adds block policies for malicious package names reported in the OpenSSF Malicious Packages dataset across PyPI, npm, Cargo, Maven, NuGet, and RubyGems.
   - Also keeps a separate builtin for package index override blocking.

5. Install-aware command normalization
   - Extends shell command normalization with a new `install` action.
   - Detects common install workflows such as `pip install`, `python -m pip install`, `uv pip install`, `npm install`, `pnpm add`, `yarn add`, `cargo add`, `dotnet add package`, `gem install`, `bundle add`, and related forms.
   - Tightens malicious package policies so they only trigger on installation-style commands instead of all shell commands.

6. Frontend guardrails UX improvements for large policy lists
   - Prevents long match summaries from overflowing the Policies list UI.
   - Avoids freezing the Rules page when a policy contains very large pattern/resource/term lists by switching oversized lists to a preview-only mode while preserving the full list on save.

**Threat Model Mapping**

This PR is intended to reduce exposure to the following paper-style attack patterns:

- Response-side payload injection into shell/bootstrap commands
- Dependency-targeted package substitution during install flows
- Passive request-side secret collection via router plaintext visibility
- Known-malicious fetch destinations and known-malicious package names appearing in tool-call arguments

This does not provide end-to-end response provenance or cryptographic integrity. It is a practical client-side guardrail layer intended to reduce risk today.

**Why This Helps**

The paper’s main observation is that a malicious or compromised API router can alter tool-call payloads after model inference and can also observe secrets in plaintext. That means schema validation alone is insufficient.

This PR adds defenses at the places we can enforce locally:
- before risky shell commands run
- before sensitive local files are read
- before secrets leave the workstation
- before known malicious packages are installed

**Limitations**

- Package blocklists are still term-based policies, not full installer-native dependency parsers.
- These controls reduce risk but do not prove that a router preserved the upstream model response.
- Some ecosystems with very large malicious package datasets rely on large builtin lists, which is why the frontend changes are included in the same PR.

**Follow-up Work**

- Add richer installer-aware matching so policies can express installer + package constraints more precisely
- Add support for external list references instead of embedding large builtin datasets directly in YAML
- Add router trust-aware policy modes for stricter enforcement on untrusted endpoints
- Explore transparency logging / provenance surfaces for stronger forensics


internal/guardrails/command_normalized.go
```diff
@@ -121,6 +124,108 @@ func normalizeShellAction(program string) string {
 	}
 }
 
+func isInstallCommand(cmd ShellSimpleCommand) bool {
```
richardxx, line 127: **Important: Missing Test Coverage**

This `isInstallCommand` function is security-critical - it determines whether package installation policies are applied. However, there are no unit tests for this function.

**Required test coverage:**
- Positive test cases for each of the 15+ package managers
- Negative test cases (non-install commands)
- Edge cases (flags, option ordering, mixed case)
- Tests for helper functions (firstArgIn, hasArgSequence, containsArg)

Without tests, we cannot verify this security-critical logic works correctly for all supported package managers.

internal/guardrails/command_normalized.go
```diff
@@ -121,6 +124,108 @@ func normalizeShellAction(program string) string {
...
+	case "mvn", "mvnw":
+		return firstArgIn(args, "dependency:get", "dependency:copy", "dependency:resolve")
+	case "code", "code-insiders", "codium":
+		return containsArg(args, "--install-extension")
+	case "openvsx", "ovsx":
+		return firstArgIn(args, "install")
+	case "git":
+		return firstArgIn(args, "clone")
+	default:
+		if strings.HasPrefix(program, "python") {
```
richardxx, line 166: **Consider: Extract package manager list to constant**

The 15+ package managers hardcoded here could benefit from being extracted to a configurable list or constant. This would:
- Make it easier to add new package managers
- Allow users to customize the list
- Improve testability

Consider a `[]PackageManager` constant with detection patterns.

frontend/src/pages/guardrails/RulesPage.tsx
```diff
@@ -123,6 +123,25 @@ type EditorState = {
...
+};
+
+type PreparedEditorState = {
+    state: EditorState;
+    oversized: Partial<Record<EditorListField, OversizedListField>>;
+};
+
+const MAX_SUMMARY_VALUES = 2;
+const MAX_SUMMARY_CHARS = 140;
+const MAX_INLINE_LIST_ITEMS = 400;
```
richardxx, line 155: **Important: Document Performance Constants**

These performance limit constants (`MAX_INLINE_LIST_ITEMS = 400`, `MAX_INLINE_LIST_CHARS = 16000`) appear arbitrary. Please add documentation explaining:

1. **Why these specific values?** - What performance analysis led to these numbers?
2. **What do they optimize for?** - UI responsiveness? Memory usage? Render time?
3. **How do they relate to actual data?** - With RubyGems at 977 lines and NuGet at 789 lines, how were these tested?

Consider adding benchmarks or performance tests to validate these choices.

frontend/src/pages/guardrails/RulesPage.tsx
```diff
@@ -123,6 +123,25 @@ type EditorState = {
...
+const MAX_INLINE_LIST_CHARS = 16000;
```
richardxx, line 158: **Minor: Consider making these configurable**

Hard-coded performance limits may not work well for all environments. Consider:
- Making these configurable via environment variables or settings
- Adding logic to auto-adjust based on detected performance
- Documenting the performance characteristics users can expect

internal/guardrails/command_normalized.go
```diff
@@ -121,6 +124,108 @@ func normalizeShellAction(program string) string {
...
 }
 
+func isInstallCommand(cmd ShellSimpleCommand) bool {
+	program := strings.ToLower(strings.TrimSpace(cmd.Program))
+	args := lowerTrimmed(cmd.Args)
+	if program == "" {
+		return false
+	}
+
+	switch program {
```
richardxx, line 134: **Security: Verify edge cases**

Please verify these edge cases are handled correctly:

1. **Flag handling:** Does `npm install --save package` work the same as `npm install package`?
2. **Option ordering:** Does `cargo install --debug package` work?
3. **Multi-arg commands:** Does `python -m pip install` correctly require the sequence?

These should be covered in the test suite.

richardxx (changes requested):
**Overall Assessment: Needs Work 🟡**

This is a well-structured first implementation with good code quality, but has important gaps to address before merging.

## Strengths
- Strong security architecture covering 4 key attack vectors
- Comprehensive package manager detection (15+ package managers)
- Smart frontend performance optimizations for large policy lists
- Clean Go implementation with good separation of concerns

## Key Issues Requiring Attention
1. **Missing URLhaus integration** - PR description mentions URLhaus malicious URL blocking but not implemented
2. **No test coverage for `isInstallCommand`** - Security-critical function needs comprehensive tests
3. **Hard-coded performance limits** - Constants need documentation explaining rationale
4. **Only 5 ecosystem files** - PyPI malicious packages missing (PR mentions 6)

See inline comments for specific details.

richardxx:
## Additional Review Comments

### Important: PR Description vs Implementation Mismatch

The PR description states:
> "1. Built-in policy to block known malicious fetch targets from URLhaus"

However, I do not see any URLhaus integration in the actual changes. This is a significant discrepancy.

**Action required:** Either:
- Remove URLhaus from the PR description if not yet implemented
- Implement the URLhaus integration as described
- Update the description to reflect what was actually implemented

Similarly, the description mentions "6 ecosystems" but only 5 malicious package files are present (Cargo, Maven, NuGet, RubyGems, package_index_overrides). PyPI appears to be missing.

### Minor: Filesystem Policy Considerations

**Git secret coverage:** The policy blocks  and  reads, but consider also blocking:
-  (can contain credentials)
-  in subdirectories
-  in subdirectories

Git credentials can be stored in various locations within a repository.

**Case sensitivity inconsistency:** In , path matching uses case-sensitive patterns (e.g., , ) but in , some patterns use . This inconsistency could lead to missed detections.

seviezhou:
Add 3 large files with git lfs:

```
malicious_npm_packages.yaml
malicious_pypi_packages.yaml
urlhaus_blocklist.yaml
```

internal/guardrails/command_normalized.go
seviezhou: tests added.

internal/guardrails/command_normalized.go
seviezhou: those edge cases will not affect the results, tests added.

internal/guardrails/command_normalized.go
seviezhou: Good suggestions, maybe do it in latter PRs.

frontend/src/pages/guardrails/RulesPage.tsx
seviezhou: Added comments for heuristic values.

PR #697, opened 2026-04-13
fix(probe_v2): evaluate smart routing rules before falling back to base services

andreasfoo:
## Summary

`resolveRuleTarget` in `probe_v2_handler.go` previously bypassed smart routing and always selected the first base service. This caused probes for rules like `aliyun-test` (which use smart routing to redirect away from `vmodel`) to hit the `vmodel` provider directly and receive `503 Service Unavailable`.

## Changes

**Probe V2 Smart Routing Fix:**
- `resolveRuleTarget` now evaluates smart routing rules first when `SmartEnabled` is true
- Builds a minimal scenario-appropriate request (OpenAI or Anthropic) to extract routing context
- Falls back to base services only when smart routing fails or returns no match
- Added `TestProbeV2_AliyunTestSmartRouting` to verify the fix

**Health Monitoring Enhancements:**
- Add `RateLimitedUntil` field to enforce recovery timeout window (services stay unhealthy for full `recovery_timeout_seconds` regardless of intermediate success)
- Add 1302 error code detection for GLM rate limit errors
- Fix error propagation when tokens=0 (early rate limit failures)
- Use `pipelineModeSmartAffinity` when SmartAffinity is enabled
- Add `HealthStage` to all pipelines for consistent health filtering
- Adjust log levels: health filter warnings at Warn, recovery at Info

**Routing Observability:**
- Add `X-TBE-Debug-Routing` header support for routing debug info
- Return errors from Anthropic stream for proper health tracking
- Debug headers provide: selected provider, model, routing source, matched smart rule index

## Test Plan

```bash
cd tingly-box
go test ./internal/server -run TestProbeV2_AliyunTestSmartRouting -v
```

Expected: all subtests pass, including `resolveRuleTarget_uses_smart_routing` which confirms the probe correctly routes to `glm-4.7` instead of `vmodel`.

## Behavior

- **Before**: Probe for `aliyun-test` hit `vmodel` directly (503 error)
- **After**: Probe correctly follows smart routing and reaches `glm-4.7`
- **Rate Limit Handling**: Services with 1302/429 errors stay unhealthy for 5 seconds (configurable) before recovery


FFengIll (approved):
LGTM

PR #719, opened 2026-04-17
Clean up command execution defaults and policy fragment management

seviezhou:
**Description**
This PR tightens a few recent guardrails changes around policy management and command execution defaults.

Main changes:
- move the default `command_execution` tool filter to runtime evaluator construction instead of silently writing `tool_names: [bash]` from the UI
- stop showing misleading `any tool` text in the policy list when no explicit tool filter is configured
- remove redundant `tool_names: [bash]` entries from builtin `command_execution` policies
- improve imported fragment handling for policy import/export and local custom fragment storage
- fix concurrent registry installs racing on `guardrails.yaml` updates


FFengIll (approved):
LGTM

PR #720, opened 2026-04-17
feat(server, frontend): add multi-tenant API token system with per-user data isolation

FFengIll:
## Summary

Enables multiple users/clients to access shared Tingly Box models without sharing credentials. 

Each API token provides isolated usage tracking and can be independently revoked, solving the credential sharing problem for multi-client deployments.

## Major

- **Multi-tenant authentication**: API tokens with `tb-share-` prefix authenticate independently from global model token
- **Per-user data isolation**: Each token has unique `user_uuid`; usage stats, logs, and metrics are isolated per token
- **Token management UI**: Full CRUD interface at `/tingly-box-tokens` — create, list, enable/disable, delete tokens
- **Backward compatible**: Global model token still works; can be disabled via `multi_tenant.disable_global_token` config
- **Automatic migration**: Existing configs get multi-tenant enabled by default (20260416 migration)

## Minor

- Added `APITokenStore` to `StoreManager` with SQLite persistence
- `GetUserIDFromContext()` prioritizes `user_uuid` from API tokens over `enterprise_user_id`
- Usage tracking endpoints filter by `user_uuid` context for data isolation
- Config helpers: `IsMultiTenantEnabled()`, `GetAPITokenSecret()`, `SetMultiTenantConfig()`
- Fixed Vite proxy path from `/api` to `/api/` for proper routing

image

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -35,7 +35,7 @@ export default defineConfig(({ mode }) => {
         },
         server: {
             proxy: useMock ? {} : {
-                '/api': {
+                '/api/': {
                     target: 'http://localhost:12580',
                     changeOrigin: true,
                     secure: false,
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -326,7 +326,7 @@ function AppContent() {
                     {/* Other routes */}
                     <Route path="/system" element={<System />} />
                     <Route path="/access-control" element={<AccessControl />} />
-                    <Route path="/tingly-box-tokens" element={<APITokensPage />} />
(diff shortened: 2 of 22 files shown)


frontend/vite.config.ts
```diff
@@ -35,7 +35,7 @@ export default defineConfig(({ mode }) => {
         },
         server: {
             proxy: useMock ? {} : {
-                '/api': {
+                '/api/': {
```
seviezhou, line 38: what is the purpose of adding another `/`

FFengIll: only /api/ use mock, but general route like /api-something

seviezhou (approved):

PR #778, opened 2026-04-25
feat: use request URL instead of startup host for OpenCode config generation

holywen:
## Summary

- Add `getBaseURLFromRequest` function to get base URL from HTTP request instead of using startup host parameter
- This ensures users get the URL they actually used to access the server
- Supports X-Forwarded-Proto header for reverse proxy scenarios

## Changes

- Modified: `internal/server/module/configapply/handler.go`
  - Add `getBaseURLFromRequest` helper function
  - Update `ApplyClaudeConfig`, `ApplyOpenCodeConfigFromState`, `GetOpenCodeConfigPreview` to use request URL
- Affects: configuration JSON and scripts for Windows/Linux/macOS

### DIFF
--- a/internal/server/module/configapply/handler.go
+++ b/internal/server/module/configapply/handler.go
@@ -13,6 +13,32 @@ import (
 	"github.com/tingly-dev/tingly-box/internal/typ"
 )
 
+// getBaseURLFromRequest constructs the base URL from the incoming HTTP request
+// This ensures users get the URL they actually used to access the server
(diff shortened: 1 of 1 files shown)


holywen:
I have tested it locally with a reverse proxy server, and confirmed it works now.
image

FFengIll:
Awesome, I understand that point. 
Thanks for the contribution.

FFengIll (approved):
I will approve this and mention `X-Forwarded-Proto` is a standard header with proxy.

PR #785, opened 2026-04-26
Fix OAuth callback retries and Codex reasoning input IDs

seviezhou:
**Summary**

This PR fixes issues:

1. Repeated OAuth attempts could fail if a stale callback server was still holding the fixed callback port.
2. Codex Responses forwarding could send malformed empty `input[].id` fields to the ChatGPT Codex backend and trigger request validation failures.

**Root Causes**

- OAuth retries could reuse a stale callback server that was still bound to the fixed local callback port after an earlier failed attempt.
- On the Codex Responses path, a `reasoning` input item could arrive without an `id`, then be converted through proxy-side OpenAI SDK request types where the reasoning item `id` is represented as a required string. That normalization step turned a missing `id` into `id: ""`, and the downstream ChatGPT Codex backend rejected the request.

**Changes**

### OAuth callback handling
- Reset proxy URL when no proxy is provided.
- Replace stale fixed-port callback servers instead of waiting for them to time out.
- Mark superseded OAuth sessions as failed.
- Ensure callback/session status is updated when callback handling or provider creation fails.
- Centralize callback server shutdown after sending the OAuth response.

### Codex request sanitization
- Sanitize outbound Codex Responses requests before forwarding to `chatgpt.com/backend-api/codex/responses`.
- Remove empty or invalid optional `input` item `id` fields.
- Preserve valid `call_id` values and the rest of the request structure.
- Prevent backend validation failures such as `Invalid 'input[*].id': ''`.
- Handle reasoning items whose missing `id` becomes `id: ""` during proxy-side request normalization.

### DIFF
--- a/internal/client/codex.go
+++ b/internal/client/codex.go
@@ -6,10 +6,13 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 )
+
+var codexInputIDPattern = regexp.MustCompile(`^[A-Za-z0-9_-]+$`)
 
 const reasoningMarker = "reasoning.encrypted_content"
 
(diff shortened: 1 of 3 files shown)


FFengIll (approved):
LGTM

PR #821, opened 2026-05-01
feat(protocol): add unified provider support for multiple AI protocols

0x0079:
Many AI vendors now expose both an OpenAI-compatible and an Anthropic- compatible endpoint at distinct base URLs, authenticated by the same API key. Today users had to register two separate Provider entries to use both. This change adds a fusion mode where one Provider can carry both URLs; the dispatcher picks the URL whose protocol matches the inbound client request, avoiding cross-protocol translation.

cc: #822

- ai.Provider: add optional APIBaseOpenAI / APIBaseAnthropic plus a ResolveEndpoint(clientStyle) helper. Legacy APIBase / APIStyle are untouched - single-protocol providers behave exactly as before.
- internal/server/fusion.go: shallow-copy the provider with the resolved URL/style at the dispatch entry points (chat-completions, responses, anthropic v1 / beta / count_tokens) so downstream HTTP clients and protocol-transform code keep reading APIBase/APIStyle as they always have. Smart-routing and load-balancing remain unaffected.
- DB: add api_base_openai / api_base_anthropic columns; AutoMigrate takes care of upgrades. Existing rows get empty strings (resolver short-circuits to legacy fields).
- API: extend Create/Update/Response DTOs with the two optional fields. Reject fusion combined with OAuth or Google api_style.
- ProviderFormDialog: rewrite syncProtocolsToParent so that when both protocols are checked and the picked template advertises both URLs, the form emits api_base_openai + api_base_anthropic; allow toggling in edit mode (was disabled); hide checkboxes for OAuth providers.
- CredentialPage: stop creating two separate providers when both protocols are picked - emit a single fusion provider instead. Edit flow now propagates the new fields.
- ApiKeyTable: surface a "Fusion" chip next to the api_style badge when both fusion URLs are populated.
- Compatibility: we use a global flag `fusion_provider` to control fusion feature, default: false as disabled.

Out of scope: APIStyleGoogle, OAuth fusion, smart-routing rule changes, probe endpoint, model-list endpoint.


FFengIll (approved):
Follow #821

PR #855, opened 2026-05-06
Refactor MCP mixed tool routing and fix advisor continuation handling

andreasfoo:
## Summary

This PR reorganizes the MCP mixed-tool routing path used by Claude Code integrations and fixes the concrete advisor-related failures we saw on the Anthropic / Claude Code route.

At a high level, this branch does five things:

1. **Moves mixed tool handling onto a continuation-store model** instead of the older pending-bridge / virtual-result-merge flow
2. **Makes advisor behave as a normal virtual MCP tool** across same-format and cross-format routes
3. **Guards against duplicate native advisor injection** when the request already carries Claude Code native advisor semantics
4. **Fixes the actual live failure in Anthropic mixed callbacks** by merging continuations into the callback request with the correct message shape
5. **Keeps builtin advisor disabled by default** until the user explicitly enables it in the frontend

This PR is not a single isolated patch. It is the coherent end state of the advisor + mixed-tool routing work on this branch.

---

## Problem statement

We need `tingly/cc` to support three behaviors at the same time:

- allow tingly-box to execute selected MCP tools server-side,
- let Claude Code continue to execute its own native tools such as `Bash`, `Glob`, `Read`, etc,
- preserve a valid streaming client experience without forcing Claude Code into an invalid continuation or fallback path.

The difficult case is **mixed tool selection**:
- upstream chooses a virtual MCP tool such as advisor, and
- upstream also chooses a client-native Claude Code tool such as `Bash`

In that case, tingly-box must execute the virtual tool itself, but still hand the native tool back to Claude Code and later stitch both results back into the next upstream request correctly.

---

### 1) Replaced the old mixed-result bridge with a continuation-store model

Main files:
- `internal/server/module/mcp/continuation_store.go`
- `internal/server/module/mcp/generic_stream_interceptor.go`
- `internal/server/module/mcp/generic_loop_processor.go`
- `internal/server/protocol_dispatch.go`
- `internal/server/mcp_pending_bridge.go` (removed)
- `internal/server/mcp_virtual_result_merge.go` (removed)
- `internal/server/module/mcp/pending_results_manager.go` (removed)

Before:
- mixed-path handling relied on pending-result bridging / anchor-based reinjection logic
- that made stream/non-stream behavior harder to reason about
- it also made cross-format continuation handling fragile

After:
- mixed virtual-tool results are stored as a continuation segment keyed by session/provider/adapter
- the next matching client callback consumes that continuation exactly once
- the same idea is used consistently in both streaming and non-streaming generic MCP paths

Why this matters:
- it gives us one uniform model for `PureVirtual`, `PureExternal`, and `Mixed`
- it removes special-case “bridge” logic that was hard to maintain and easy to break during cross-format conversion

---

### 2) Normalized advisor as a standard virtual MCP tool

Main files:
- `internal/server/module/mcp/tool_classification.go`
- `internal/server/module/mcp/tool_executor.go`
- `internal/server/module/mcp/message_extract.go`
- `internal/server/module/mcp/openai_chat_adapter.go`
- `internal/server/module/mcp/format_adapter.go`
- `internal/server/mcp_tool_error.go`
- `internal/server/transform/transform_mcp_tool_injection.go`

Behavioral changes:
- advisor is treated like any other virtual MCP tool once it is injected
- classification is name-based and registry-based
- non-MCP tool names such as `Bash` are kept out of the MCP runtime path
- execution paths explicitly recognize MCP-prefixed tools and avoid accidentally routing client-native tools through MCP execution

Why this matters:
- advisor does **not** need to masquerade as Claude Code native `server_tool_use`
- it can coexist with client-native tools as long as mixed routing and continuation replay are correct

---

### 3) Added native advisor duplicate-injection guard

Main files:
- `internal/server/native_advisor_detection.go`
- `internal/server/transform/transform_mcp_tool_injection.go`
- `internal/server/transform/transform_mcp_tool_injection_test.go`
- `internal/server/config/config.go`

Behavioral changes:
- when a request already contains Claude Code native advisor semantics, tingly-box does not inject its own advisor MCP tool again
- this avoids double-advisor situations for users who already have a native advisor path enabled

Why this matters:
- prevents duplicate advisor behavior without forcing the tingly-box advisor itself to become a native Claude Code server tool
- keeps the injected advisor path safe for users with different Claude-side setups

---

### 4) Fixed cross-format advisor continuation handling

Main files:
- `internal/protocol/stream/anthropic_to_openai.go`
- `internal/protocol/stream/openai_to_anthropic.go`
- `internal/protocol/stream/openai_to_anthropic_beta.go`
- `internal/server/anthropic_to_openai_mcp_stream.go`
- `internal/server/openai_to_anthropic_mcp_stream.go`
- `internal/server/protocol_transform.go`
- `internal/server/cross_format_test.go`
- `internal/server/mcp_anthropic_v1_dispatch_test.go`
- `internal/server/mcp_anthropic_v1_helper.go`

Behavioral changes:
- same-format and cross-format routes now carry virtual tool results through the generic continuation path instead of relying on the old bridge logic
- advisor can continue its server-side loop even when protocol conversion is involved

Why this matters:
- advisor support is not limited to Anthropic→Anthropic
- it also works across OpenAI↔Anthropic routing paths in the generic MCP design

---

### 5) Fixed remote MCP runtime session lifetime for HTTP/SSE sources

Main files:
- `internal/mcp/runtime/session.go`
- `internal/mcp/runtime/source_http.go`
- `internal/mcp/runtime/source_sse.go`
- `internal/server/mcp_remote_runtime_e2e_test.go`

Behavioral changes:
- stop using a request-scoped / timeout context as the MCP SDK connection lifetime for long-lived remote sources
- use a non-canceling context for `Connect(...)`
- remove redundant source-level locking around remote source tool calls and listing
- add HTTP and SSE remote runtime regression coverage

Why this matters:
- remote MCP sessions were at risk of being torn down too early because the SDK binds transport lifetime to the context passed to `Connect`
- this fix makes remote MCP HTTP/SSE sources stable in the normal runtime path

---

### 6) Keep builtin advisor disabled by default until explicitly enabled

Main files:
- `frontend/src/pages/mcp/MCPRegisteredServers.tsx`
- `internal/mcp/runtime/runtime.go`
- `internal/mcp/runtime/runtime_test.go`

Behavioral changes:
- frontend now shows builtin advisor as disabled by default when no prior config exists
- more importantly, runtime no longer treats advisor as implicitly enabled just because the virtual tool is registered
- advisor is only exposed in enabled server-tool names when the advisor MCP source is explicitly enabled in config

Why this matters:
- a user who has never enabled MCP/advisor should not get advisor behavior by default
- this makes the default state consistent across UI, config semantics, and server-side execution paths

---

## The actual live root cause we found

The most important part of this branch is the final fix for the real Claude Code mixed-flow failure.

### Symptom

On the Anthropic Beta / Claude Code route, a mixed response could succeed for the first round, but fail on the next callback after Claude Code returned the external tool result.

The upstream provider then rejected the follow-up request with an Anthropic error equivalent to:

- `tool_use ids were found without tool_result blocks immediately after ...`

### What the bug was

The bug was **not**:
- that advisor was a normal MCP tool
- that advisor needed to be rewritten into Claude Code native `server_tool_use`
- that mixed mode itself was conceptually wrong

The bug **was**:
- the stored Anthropic continuation was being applied too naively
- the old logic effectively prefixed the stashed continuation in front of the callback request messages

That is wrong in live mixed callback shapes, because Claude Code often sends back something like:

1. leading user prompt / wrapper text
2. `assistant(tool_use external)`
3. `user(tool_result external)`

Blindly prepending the stored continuation creates an invalid message sequence and breaks Anthropic's required adjacency between `tool_use` and the corresponding `tool_result`.

### The fix

Main files:
- `internal/server/module/mcp/anthropic_beta_adapter.go`
- `internal/server/module/mcp/anthropic_v1_adapter.go`
- `internal/server/module/mcp/anthropic_continuation_merge_test.go`
- `internal/server/module/mcp/anthropic_continuation_realshape_test.go`
- `internal/server/mcp_path_matrix_e2e_test.go`

New behavior:
- if the callback contains only a `user(tool_result)`, merge the virtual result into that user message
- if the callback contains a full `assistant(tool_use) -> user(tool_result)` pair, replace that pair with the stashed mixed continuation and merge the external `tool_result` into the continuation's user result message
- preserve leading user prompt / wrapper messages in front of the merged continuation

Correct follow-up shape becomes:

1. user prompt / wrapper text
2. assistant `tool_use(advisor + external)`
3. user `tool_result(advisor + external)`

This is the shape the upstream Anthropic route expects.

---

### Real black-box validation

We validated the final mixed path with a real Claude Code run against local `tingly/cc`.

Scenario:
- upstream selected both:
  - the virtual advisor MCP tool
  - client-native `Bash`
- Claude Code executed `Bash`
- tingly-box replayed the stored continuation on the callback
- upstream accepted the merged request shape
- final result completed successfully as:

```text
MIXED_OK
```

### Tests / checks

Ran:
- `go test ./internal/server/module/mcp -run 'TestAnthropic(Beta|V1)ApplyContinuation_Merges(IntoFirstUserMessageEvenWithLeadingAssistantOrText|IncomingToolResultIntoSameUserMessage)'`
- `go test ./internal/server -run 'TestMCPMixedToolStreamStashInjectE2E|TestMCPPathMatrixE2E'`
- `go test ./internal/mcp/runtime -run 'TestRegisterBuiltinTools|TestListEnabledServerToolNames'`
- `task cli:build`

---

## Notes for reviewers

If you want to review this branch efficiently, the best order is:

1. `continuation_store.go` + generic loop/interceptor changes
2. adapter-level Anthropic continuation merge logic
3. native advisor duplicate-injection guard
4. runtime enabled-name gating for advisor default-off behavior
5. remote runtime session lifecycle fix
6. E2E / regression tests

The most important outcomes are:
- mixed advisor + client-native tool flows now work with advisor remaining a **normal virtual MCP tool**,
- the live Anthropic callback shape is now structurally valid,
- and builtin advisor stays **off by default** until a user explicitly enables it.


FFengIll (approved):
I will approve this since MCP is an experimental feature and we will continue to improve it.

PR #860, opened 2026-05-07
Fix guardrails command matching and blocked tool-use stream handling

seviezhou:
**Description**
This PR fixes a few guardrails regressions caused by mismatches between normalized command semantics, stream rewriting, and UI defaults.

Root causes:
- `command_execution` term matching only checked individual normalized terms, so multi-token patterns like `rm -rf` failed once commands were split into separate tokens
- blocked Anthropic `tool_use` streams were rewritten into text blocks, but the trailing `message_delta.stop_reason` still remained `tool_use`, which left the client in an inconsistent state and could trigger duplicate block output or tool-call parse failures
- the UI was implicitly persisting `tool_names: [bash]` for `command_execution` policies even though that filter was no longer clearly exposed in the editor

Changes:
- fix `command_execution` term matching so multi-token patterns match the joined normalized command sequence
- rewrite blocked Anthropic `tool_use` streams so the following `message_delta.stop_reason` becomes `end_turn`
- move the default `command_execution` tool filter to runtime evaluator construction
- update tests to match the current normalized command action model

### DIFF
--- a/internal/guardrails/core/command_shell_test.go
+++ b/internal/guardrails/core/command_shell_test.go
@@ -114,7 +114,7 @@ func TestCommandAttachDerivedFieldsBuildsShellAndNormalizedViews(t *testing.T) {
 	if cmd.Normalized.Kind != "shell" {
 		t.Fatalf("expected shell kind, got %q", cmd.Normalized.Kind)
 	}
-	if !reflect.DeepEqual(cmd.Normalized.Actions, []string{"read"}) {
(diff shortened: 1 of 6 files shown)


FFengIll (commented):
Somehow it is hard to understand the root cause from code.
Pls provide a better analysis record in PR content.

seviezhou:
updated

FFengIll (approved):

PR #861, opened 2026-05-07
Fix scenario V2 recording activation and add three recording modes

andreasfoo:
## Summary
- make scenario-level `recording_v2` actually activate V2 staged recording for flows such as `claude_code`
- route staged request/response capture through `ProtocolRecorder` consistently, instead of falling back to legacy scenario recording in some Anthropic paths
- avoid mixing legacy roundtrip JSONL output into V2 scenario recording modes
- replace the old `request/response/both` shape with three recording modes: `request`, `request_response`, and `staged_request_response`
- update the frontend mode selector to reflect the new recording semantics

## Problem
Scenario config already supported `recording_v2`, but `claude_code` requests still did not emit true staged V2 entries (`original_request`, `transformed_request`, `final_response`). In practice this made it hard to determine whether malformed tool calls came from the upstream client/LLM or from tingly-box transforms.

## Root Cause
There were multiple gaps in the recording path:
- scenario sinks were still created with legacy `scenario` mode instead of the scenario’s effective `recording_v2` mode
- several Anthropic handlers still passed global `s.recordMode` instead of the scenario-effective mode
- transform pre/post recorders were gated behind a stale global `enableRecording` path that was not what scenario `recording_v2` actually used
- client roundtrip recording could still write old-format JSONL alongside V2 staged entries, making inspection noisy
- the old recording mode model did not clearly separate final-request recording from staged request/transform recording

## Fix
- compute recording mode from the scenario via `GetScenarioRecordMode(...)`
- create scenario sinks with that effective mode
- pass the scenario-effective mode into V2 recorder setup and transform-chain recording decisions
- keep non-stream Responses→Anthropic paths on `ProtocolRecorder`
- skip legacy roundtrip recording when V2 scenario recording is active
- redefine scenario `recording_v2` into three modes:
  - `request`: final outbound request only
  - `request_response`: final outbound request + final response
  - `staged_request_response`: original request + transformed request + final response
- update frontend scenario controls to use the new three-mode semantics

## Validation
- rebased onto `origin/main`
- `task cli:build` passes in the isolated worktree
- verified on a worktree build running on port `12580`
- switched `claude_code.recording_v2` through `request`, `request_response`, and `staged_request_response`
- sent real `claude_code` requests against the patched server
- confirmed the appended JSONL entries under `~/.tingly-box-daily/record/claude_code` match the expected field sets for each mode

### DIFF
--- a/frontend/src/components/ExperimentalFeatures.tsx
+++ b/frontend/src/components/ExperimentalFeatures.tsx
@@ -32,9 +32,9 @@ const FEATURES: FeatureConfig[] = [
 // Record V2 modes
 const RECORD_V2_MODES = [
     { value: '', label: 'Off', description: 'Recording disabled' },
-    { value: 'request', label: 'Request', description: 'Record request only' },
-    { value: 'response', label: 'Response', description: 'Record response only' },
-    { value: 'request_response', label: 'Both', description: 'Record both request and response' },
+    { value: 'request', label: 'Request Only', description: 'Record the final outbound request only' },
+    { value: 'request_response', label: 'Request + Response', description: 'Record the final outbound request and final response' },
+    { value: 'staged_request_response', label: 'Request + Transform + Response', description: 'Record original request, transformed request, and final response' },
 ] as const;
 
 const ExperimentalFeatures: React.FC<ExperimentalFeaturesProps> = ({ scenario }) => {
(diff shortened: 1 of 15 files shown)


FFengIll (approved):
LGTM, and hope it help.

PR #867, opened 2026-05-07
Fix Codex probe requests to use the Responses API and Codex-compatible tool schemas

seviezhou:
**Summary**

This PR fixes probe failures for Codex-backed providers and rules.

When a probe targeted a provider or rule that ultimately used the ChatGPT Codex backend, the probe flow could still send Chat Completions style requests with `messages`, use non-stream Responses calls against an SSE endpoint, or send tool schemas that were valid for standard OpenAI chat but rejected by the Codex backend. These mismatches caused probe errors even though the underlying provider configuration was otherwise correct.

截屏2026-05-07 17 30 06

**Root Cause**

- Probe V2 SDK paths built `/chat/completions` requests for OpenAI-style targets without accounting for providers that actually route to the ChatGPT Codex backend.
- Codex backend probe requests need to use the Responses API and, in practice, stream mode semantics.
- Codex backend also validates function tool schemas more strictly and requires `additionalProperties: false` in the probe function schema.

**Changes**

### Probe V2 Responses support
- Add Responses API request construction for Probe V2 SDK flows.
- Route Codex-backed OpenAI probes to `/responses` instead of `/chat/completions`.
- Apply the same Codex-aware behavior to simple, streaming, and tool probe modes.

### Streaming handling for Codex probes
- Use `ResponsesNewStreaming()` for Codex-backed simple probes as well as streaming probes.
- Collect streamed events and return them as probe output instead of attempting a non-stream Responses call against an SSE response.

### Codex-compatible tool schemas
- Update probe function tool schemas for OpenAI/Responses probe requests to include:
  - `type: "object"`
  - `additionalProperties: false`
- Keep the existing probe tool shape while making it acceptable to the Codex backend.

### DIFF
--- a/frontend/src/components/CodexConfigModal.tsx
+++ b/frontend/src/components/CodexConfigModal.tsx
@@ -38,6 +38,8 @@ const CodexConfigModal: React.FC<CodexConfigModalProps> = ({
 
     const configToml = `model = "tingly-codex"
 model_provider = "tingly-box"
+model_supports_reasoning_summaries = true
+model_reasoning_summary = "auto"
 
 [model_providers.tingly-box]
 name = "OpenAI using Tingly Box"
--- a/frontend/src/pages/scenario/UseCodexPage.tsx
+++ b/frontend/src/pages/scenario/UseCodexPage.tsx
@@ -27,7 +27,7 @@ const UseCodexPageContent: React.FC = () => {
     // Codex has no backend apply — copy config.toml to clipboard as a convenience
     const handleApply = async (): Promise<AgentApplyResult> => {
         const codexBaseUrl = `${baseUrl}/tingly/codex`;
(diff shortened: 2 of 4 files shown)


FFengIll:
Tingly Box provides 3 probe testing - nonstream, stream and tool (in stream).

I consider to merge stream and tool into one - stream (with tool).

seviezhou:
This pr currently fixes all 3 probe testing methods.

FFengIll:
> This pr currently fixes all 3 probe testing methods.

I mean we can remove 1, and keep only 2 in frontend action.

seviezhou:
Ok, I think we can do it. Should we merge this pr first or later?

FFengIll:
Merge this first is ok. Approved.

FFengIll (approved):

PR #875, opened 2026-05-08
fix(transform): merge consecutive function_call items and align tool messages for DeepSeek compatibility

Chee-0806:
## Summary

Fixes 400 errors from DeepSeek when using tool calls:
`An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'`
and
`The reasoning_content in the thinking mode must be passed back to the API.`

### 1. Consecutive function_call items → separate assistant messages (Responses→Chat)

`ConvertResponsesInputToMessages` created one assistant message per `function_call` input item, each with a single `tool_calls` entry. This produced sequences like:

```
assistant(tool_calls: [call_0])
assistant(tool_calls: [call_1])     ← DeepSeek rejects: assistant(call_0) not immediately followed by tool messages
tool(call_0)
tool(call_1)
```

**Fix**: Merge consecutive `function_call` items into a single assistant message with all `tool_calls`.

### 2. Premature flush on `message` items (Responses→Chat)

When Responses API input has the order `function_call → message → function_call_output`, `flushCalls` was called on encountering a `message` item, producing:

```
assistant(tool_calls) → user → tool
```

DeepSeek rejects this because assistant messages with `tool_calls` must be immediately followed by their tool responses.

**Fix**: Remove the `flushCalls` call on `message` items. Pending function_calls are now only flushed when a `function_call_output` is encountered.

### 3. `reasoning_content` not serialized (DeepSeek thinking mode)

`ChatCompletionMessageParamUnion` is a wrapper around concrete variants (`OfAssistant`, `OfUser`, `OfTool`). When `OfAssistant` is set, the SDK marshals the variant itself — union-level `ExtraFields` are dropped. The DeepSeek transform was writing `reasoning_content` to the union level, so it never appeared in the outbound JSON.

**Fix**: Write `reasoning_content` (and `x_thinking` during Anthropic→OpenAI conversion) to `OfAssistant.SetExtraFields()`, which is the object actually serialized.

## Changes by file

| File | Change | Scope |
|------|--------|-------|
| `openai_responses_to_chat.go` | Merge consecutive `function_call` items into one assistant message; defer flush until `function_call_output` | Universal |
| `request_openai_deepseek.go` | Write `reasoning_content` to `OfAssistant.ExtraFields` (not union level); skip empty overwrite | DeepSeek/Moonshot |
| `request_openai_deepseek_test.go` | Add serialization behavior tests (union vs variant extras, x_thinking→reasoning_content) | DeepSeek/Moonshot |
| `request_openai_extensions.go` | Register DeepSeek via `ProviderConfigs` instead of model-name fallback | DeepSeek |
| `anthropic_beta_to_openai.go` | Store `x_thinking` on `OfAssistant` variant level | Universal |
| `anthropic_v1_to_openai.go` | Store `x_thinking` on `OfAssistant` variant level | Universal |
| `thinking_helper.go` | Accept union-level response for compatible callers | Universal |
| `anthropic_to_openai.go` | Guard empty `reasoning_content` in streaming chunks | Universal |
| `consistency.go` | `AlignToolMessagesForOpenAI` — single-pass orphaned tool message conversion only (no dual-pass) | Universal |
| `chain_builder.go` | Add `ConsistencyTransform` after MCP transforms, before VendorTransform | Universal |
| `protocol_dispatch.go` | Remove redundant `AlignToolMessagesForOpenAI` calls | Universal |

## Discussion summary

- seviezhou's draft PR (#887) raised the `reasoning_content` serialization issue and `consistency.go` concerns — both addressed
- `consistency.go` dual-pass reverted to single-pass (the root cause fix makes defense-in-depth unnecessary)
- SDK serialization tests (`request_openai_deepseek_test.go`) added to verify union vs variant ExtraFields behavior

## Test plan

- [x] Existing tests pass (`go test ./internal/protocol/...`)
- [x] Manually verified with Codex + DeepSeek via opencode.ai proxy
- [x] SDK serialization tests verify union-level vs variant-level ExtraFields
- [x] Message alignment is idempotent (multiple passes don't corrupt valid sequences)


FFengIll:
Please do a double check @andreasfoo @seviezhou

FFengIll:
@Chee-0806 Thanks for your contribution.

BTW, I see `internal/protocol/transform/consistency.go` is changed too, is there any special reason?

Chee-0806:
The change to `consistency.go` expands the `AlignToolMessagesForOpenAI` function from a single-pass cleanup to a two-pass safety net:

**Before**: It only handled one direction — converting orphaned **tool messages** (those whose `tool_call_id` has no matching assistant `tool_calls`) into user messages.

**Now**: It also handles the reverse direction — **Pass 1** strips `tool_calls` from assistant messages that have no corresponding tool messages in the history. This prevents errors like `"An assistant message with tool_calls must be followed by tool messages"` from providers that enforce strict ordering (not just DeepSeek).

The pass order is deliberate:
1. **Pass 1**: Strip orphaned tool_calls first
2. **Pass 2**: Then convert orphaned tool messages (re-collecting valid IDs after Pass 1)

If we ran Pass 2 first, then stripped tool_calls in Pass 1, the stripped IDs might belong to tool messages that Pass 2 had already validated — creating new undetected orphans.

The primary fix is in `openai_responses_to_chat.go` (merging consecutive `function_call` items into one assistant message), but this consistency transform serves as a defense-in-depth layer — catching any edge case from other code paths that might produce invalid message sequences.

internal/protocol/request/openai_responses_to_chat.go
FFengIll, line 1: Agree to merge parallel tool call into list of call.

internal/protocol/stream/anthropic_to_openai.go
```diff
@@ -400,6 +400,12 @@ func createReasoningContentChunk(chatID string, created int64, model, reasoning
 		},
 	}
 
+	// Only add reasoning_content if reasoning is not empty
+	// DeepSeek and other providers reject empty reasoning_content fields
+	if reasoning == "" {
```
FFengIll, line 405: I agree, and it makes sense.

internal/protocol/transform/ops/request_openai_deepseek.go
```diff
@@ -25,23 +22,20 @@ func applyDeepSeekTransform(req *openai.ChatCompletionNewParams, providerURL, mo
...
-				// Ensure reasoning_content field exists even if no thinking content
-				// Use a placeholder (empty pointer) instead of empty string to ensure it's included in JSON
-				var emptyStr string
-				msgMap["reasoning_content"] = &emptyStr
+			} else if _, hasReasoning := msgMap["reasoning_content"]; !hasReasoning {
+				// DeepSeek requires reasoning_content on assistant messages, especially
+				// those with tool_calls. Per DeepSeek docs: "For turns that do perform
+				// tool calls, the reasoning_content must be fully passed back to the API
+				// in all subsequent requests."
+				msgMap["reasoning_content"] = ""
```
FFengIll, line 35: here might be a refactoring, it is ok.

FFengIll (commented):
Waiting for reproduction and validation @seviezhou

internal/protocol/transform/consistency.go
```diff
@@ -223,37 +223,98 @@ func (t *ConsistencyTransform) normalizeMessages(req *openai.ChatCompletionNewPa
 }
 
 // AlignToolMessagesForOpenAI converts orphaned tool messages (those without a
-// matching tool_call_id) to user messages. This prevents "role 'tool' must be a
-// response to preceding message with 'tool_calls'" errors.
+// matching tool_call_id) to user messages, and strips tool_calls from assistant
+// messages that lack corresponding tool messages. This prevents API errors like:
+// "An assistant message with 'tool_calls' must be followed by tool messages
+// responding to each 'tool_call_id'."
```
0x0079: I think the consistency change is not necessary, may you revert this and test again?

seviezhou:
@Chee-0806 Thanks for the fix, after verifying your code, I think most of them are necessary and correct. Moreover, I think we might remove some unnecessary parts and refine some changes, such as changes to `consistency.go`, I wrote some discussions in a draft pr (#887 ), maybe we can discuss more about it so that you can further refine your PR.

Chee-0806:
@seviezhou Thanks for the thorough review and the analysis in #887! I've reviewed your draft PR and have a few thoughts.

## consistency.go

Agreed — I've reverted the dual-pass changes back to the original single-pass (orphaned tool message conversion only). The root cause fix (merging consecutive function_calls) makes Pass 1 unnecessary. Thanks to you and @0x0079 for flagging this.

## Regarding reasoning_content write location

Your draft PR mentions that `reasoning_content` must be written to `OfAssistant.SetExtraFields()` because the SDK only serializes the active variant. I took a closer look at the SDK source and found the behavior is slightly different.

`ChatCompletionMessageParamUnion.MarshalJSON()` actually checks for union-level ExtraFields and, when present, calls `MarshalWithExtras`, which merges them into the variant's JSON output via `sjson.SetBytes`:

```go
// chatcompletion.go:2048-2081
func (u ChatCompletionMessageParamUnion) MarshalJSON() ([]byte, error) {
    extras := u.ExtraFields()
    if len(extras) > 0 {
        variant = u.OfAssistant  // find active variant
        return param.MarshalWithExtras(u, variant, extras)  // merges extras in
    }
    return param.MarshalUnion(u, ...)  // fallback: just marshal variant
}
```

So **union-level ExtraFields do appear in the final JSON**. I added a set of tests to verify this behavior end-to-end:

| Test | What it verifies |
|------|------------------|
| `TestUnionLevelExtraFieldsSerialized` | Union-level `reasoning_content` appears in JSON |
| `TestOfAssistantLevelExtraFieldsSerialized` | Variant-level extras also appear |
| `TestUnionExtrasOverrideVariantExtras` | Union extras win on key overlap |
| `TestDeepSeekTransformReadsUnionLevelXThinking` | `x_thinking` → `reasoning_content` works at union level |
| `TestDeepSeekTransformFallbackEmptyReasoningContent` | Empty string fallback serializes correctly |
| `TestAnthropicConversionStoresXThinkingOnUnionLevel` | Anthropic→OpenAI stores `x_thinking` at union level |

All 6 tests pass. The current PR's approach of writing to union level works correctly.

That said, your approach of checking both levels (union + OfAssistant) in #887 is more defensive. If the team prefers that style, I'm happy to adopt it.

## Summary of changes in this update

1. ✅ Reverted `consistency.go` dual-pass → single-pass (your + @0x0079's suggestion)
2. ✅ Added SDK serialization tests as verification
3. ✅ Synced fork with upstream main

Thanks again for the careful review!

internal/protocol/transform/consistency.go
Chee-0806: @0x0079 Good catch — I've reverted the dual-pass changes in `AlignToolMessagesForOpenAI` back to the original single-pass (orphaned tool message conversion only). The root cause fix (merging consecutive function_calls in `openai_responses_to_chat.go`) makes the extra pass unnecessary. Thanks!

Chee-0806:
## Additional fix during testing

While testing the PR changes, I hit another DeepSeek 400 error with the same `insufficient tool messages following tool_calls message` message. Root cause:

`ConvertResponsesInputToMessages` was calling `flushCalls(pendingCalls)` when encountering a `message` item. When the Responses API input has the order `function_call → message → function_call_output`, this produces an invalid message sequence:

```
assistant(tool_calls) → user → tool
```

DeepSeek rejects this because assistant messages with `tool_calls` must be immediately followed by their tool responses.

**Fix**: Remove the `flushCalls` call on `message` items. Pending function_calls are now only flushed when a `function_call_output` is encountered, ensuring `assistant(tool_calls)` always appears directly before its tool messages.

This was a pre-existing issue in the original code (not introduced by this PR), but it was never triggered because the Responses API input previously didn't contain this edge-case ordering.

internal/protocol/transform/ops/request_openai_extensions.go
```diff
@@ -65,23 +65,27 @@ var ProviderConfigs = []providerConfig{
...
+					return config.Transform
+				}
 			}
 		}
 	}
 
+	// Fallback: detect DeepSeek models by model name for proxies/aggregators
+	// DeepSeek reasoning models (deepseek-reasoner, deepseek-r1, deepseek-r1-*)
+	// and DeepSeek V-series thinking models require reasoning_content handling
+	if strings.Contains(modelLower, "deepseek") {
```
seviezhou: Can we change to use `ProviderConfigs` to apply the transformation? Like here: [https://github.com/tingly-dev/tingly-box/pull/887/changes#diff-4d46a45e3f7c1a7349413c002804e6e60e5b9863611a5b63a77034789da8be0cR31](https://github.com/tingly-dev/tingly-box/pull/887/changes#diff-4d46a45e3f7c1a7349413c002804e6e60e5b9863611a5b63a77034789da8be0cR31)

Chee-0806: Agreed, the model-name fallback is a magic string and doesn't follow engineering best practices. Removed it and added opencode.ai/zen/go to ProviderConfigs instead. See the latest commit on the fork branch.

internal/server/chain_builder.go
```diff
@@ -39,6 +39,8 @@ func (s *Server) BuildTransformChain(c *gin.Context, targetType protocol.APIType
 		transforms = append(transforms, servertransform.NewMCPToolInjectionTransform(s.mcpRuntime))
 		transforms = append(transforms, servertransform.NewMCPToolStripGuardTransform(s.mcpRuntime, s.mcpStripDisabledToolsEnabled()))
 	}
+	// 3. Consistency transform (cross-provider normalization including message alignment)
+	transforms = append(transforms, transform.NewConsistencyTransform(targetType))
```
seviezhou, line 43: Is it still necessary?

internal/server/protocol_dispatch.go
```diff
@@ -644,6 +644,7 @@ func (s *Server) dispatchOpenAIChat(
 	if seg, ok := mcp.PopOpenAIContinuationSegment(typ.GetSessionID(c.Request.Context()), provider.UUID); ok {
 		req.Messages = append(append([]openai.ChatCompletionMessageParamUnion{}, seg...), req.Messages...)
 	}
+	transform.AlignToolMessagesForOpenAI(req)
 	request.CleanupOpenaiFields(req)
```
seviezhou, line 650: Is it still necessary?

internal/server/chain_builder.go
Chee-0806: Yes, it's still necessary. ConsistencyTransform handles four responsibilities that aren't covered elsewhere:

- **Tool schema normalization** — enforces type/object and cleans up empty parameters
- **Scenario flags injection** — disables stream usage, stores flags for downstream use
- **tool_call_id truncation** — enforces the 40-character OpenAI API limit
- **Request parameter validation** — validates temperature, max_tokens, top_p ranges

Only the internal `alignToolMessages` call overlaps with dispatch, but that's a minor redundancy rather than a reason to remove the entire transform.

internal/server/protocol_dispatch.go
Chee-0806: Good catch. `AlignToolMessagesForOpenAI` was redundant — the transform chain's ConsistencyTransform already calls it via `normalizeMessages → alignToolMessages` before dispatch runs for all TypeOpenAIChat targets. Removed in 47d4ce4.

FFengIll:
I've reviewed all changes at this point and would like to approve.
I'm also requesting @seviezhou for this PR since he has been more involved.

Furthermore, it would be very nice to see the update to this PR description since we changed. @Chee-0806

seviezhou (approved):

FFengIll (approved):
LGTM after changes.
Thanks for your effort.

PR #883, opened 2026-05-09
Fix 404 error fetching Deepseek model list

seviezhou:
According to [https://api-docs.deepseek.com/api/list-models](https://api-docs.deepseek.com/api/list-models), deepseek should use [https://api.deepseek.com/models](https://api.deepseek.com/models) to fetch model list, no need to use anthropic or openai specific. Avoid the following error:

```
INFO[0121] Messages endpoint is accessible               action=fetch_models api_base="https://api.deepseek.com/anthropic" models_count=0 provider=placeholder response_time=747 success=true valid=true
ERRO[0122] Failed to fetch models from API: GET "https://api.deepseek.com/anthropic/v1/models": 404 Not Found
```

### DIFF
--- a/internal/server/config/config.go
+++ b/internal/server/config/config.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 	"sync"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


internal/server/config/config.go
```diff
@@ -1813,30 +1814,41 @@ func (c *Config) FetchAndSaveProviderModels(uid string) error {
...
-		if err == nil {
-			defer gClient.Close()
-			lister = gClient
-		}
-		apiErr = err
-	case protocol.APIStyleOpenAI:
-		fallthrough
-	default:
-		oClient, err := client.NewOpenAIClient(provider, "", typ.SessionID{})
+	if strings.Contains(strings.ToLower(strings.TrimSpace(provider.APIBase)), "api.deepseek.com") {
```
FFengIll, line 1817: I get this point.
BTW, we can update provider.json as fallback.
So I suggest to use config and keep code clean and general

seviezhou: I tried to look at provider.json, but there seems no separate field for model list, adding a new one will cause more code changes. So I fix it like it, should we add a new field?

PR #906, opened 2026-05-12
refactor(mcp): stabilize advisor runtime and native web tool routing

andreasfoo:
## Summary
- refactor advisor config to resolve providers by UUID at call time and reuse the normal provider/client infrastructure instead of carrying ad-hoc base URL and API key fields
- stabilize MCP runtime and server integration by keeping local MCP servers alive across requests, restoring matrix coverage, and cleaning duplicated advisor test setup
- strip native Claude Code web search/web fetch tool calls when MCP equivalents are available, and drop unrelated DeepSeek transform overlap from this branch

## Test plan
- [x] targeted MCP and server test paths updated for advisor provider resolution
- [x] runtime and server matrix coverage restored for both MCP-enabled and no-MCP paths
- [ ] full `task go:test`

## Notes
- removes the legacy `/api/v1/probe-models` path in favor of the provider-based flow already used elsewhere
- keeps advisor limits split between per-request consultation count and per-call token budget

### DIFF
--- a/internal/client/http.go
+++ b/internal/client/http.go
@@ -496,10 +496,6 @@ func createSessionBoundTransport(provider *typ.Provider, sessionID typ.SessionID
 				model:        model,
 				proxyURL:     provider.ProxyURL,
 			}
-		case ai.IssuerCodex:
-			return &codexRoundTripper{
-				RoundTripper: baseTransport,
-			}
 		}
 	}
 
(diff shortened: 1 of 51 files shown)


andreasfoo:
mcp front end
image

internal/client/http.go
```diff
@@ -496,6 +496,10 @@ func createSessionBoundTransport(provider *typ.Provider, sessionID typ.SessionID
 				model:        model,
 				proxyURL:     provider.ProxyURL,
 			}
+		case ai.IssuerCodex:
```
FFengIll: I believe it is deleted.

FFengIll (approved):

PR #984, opened 2026-05-18
refactor(server): move cursor_compat into the transform chain

0x0079:
`cursor_compat` was a rule flag whose request-side mutation still lived
as an inline call in the OpenAI Chat handler, plus a few dead paths
(`__tb_cursor_compat` writer with no reader, `OpenAIConfig.CursorCompat`
gated branches that never fired). This PR finishes the migration by
turning it into a real `Transform`.

- New `OpenAICursorCompatTransform`, prepended to the chain before
  `BaseTransform` (same prepend pattern `smart_compact` uses).
- `ruleExtraTransforms` split into `rulePreBaseTransforms` (inbound-shape
  mutations) and `ruleExtraTransforms` (target-shape mutations). Each
  rule flag now picks one side.
- `cursor_compat_auto` header detection folded into
  `resolveRuleFlags(c, rule)` so handlers see one effective flag.
- Dead paths removed: `applyCursorCompatFlag`, the `cursor_compat` read
  in `buildOpenAIConfigFromRequest`, `config.CursorCompat` branches in
  `ApplyProviderTransforms` / `applyDeepSeekTransform`, and the
  `OpenAIConfig.CursorCompat` field.
- `.design/rule-flags.md` updated to reflect the new pre/post-Base
  split.

## Test plan

- [x] `go test ./internal/protocol/transform/... ./internal/server/`
- [x] New tests in `openai_cursor_compat_test.go` cover Chat flatten,
  no-op on other shapes, and chain ordering.
- [x] `rule_flags_test.go` updated for the pre/post split and the
  auto-detect fold.

https://claude.ai/code/session_01PDPRsAQ2MQVEEFP6YRUd6F

### DIFF
--- a/internal/protocol/transform/ops/request_openai_extensions.go
+++ b/internal/protocol/transform/ops/request_openai_extensions.go
@@ -205,11 +205,6 @@ func applyDefaultTransform(req *openai.ChatCompletionNewParams, config *protocol
 // ApplyProviderTransforms applies provider-specific transformations
 // Falls back to default handling if no specific transform found
 func ApplyProviderTransforms(req *openai.ChatCompletionNewParams, providerURL, model string, config *protocol.OpenAIConfig) *openai.ChatCompletionNewParams {
-	// When cursor_compat is enabled, ALWAYS flatten content regardless of provider
-	// This ensures Cursor compatibility for all providers
-	if config.CursorCompat {
-		normalizeCursorContent(req)
-	}
 	if transform := GetProviderTransform(providerURL, model); transform != nil {
 		return transform(req, providerURL, model, config)
 	}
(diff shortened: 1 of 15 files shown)


FFengIll (approved):

internal/protocol/transform/ops/request_openai_extensions.go
```diff
@@ -205,11 +205,6 @@ func applyDefaultTransform(req *openai.ChatCompletionNewParams, config *protocol
 // ApplyProviderTransforms applies provider-specific transformations
 // Falls back to default handling if no specific transform found
 func ApplyProviderTransforms(req *openai.ChatCompletionNewParams, providerURL, model string, config *protocol.OpenAIConfig) *openai.ChatCompletionNewParams {
-	// When cursor_compat is enabled, ALWAYS flatten content regardless of provider
```
FFengIll, line 208: 这是不当的设计，本质是为了确保openai请求的格式标准，以支持cursor，在Pre-Base-Chain操作更合理。

FFengIll (approved):

PR #992, opened 2026-05-19
fix(anthropic): preserve thinking alongside context_management for claude-opus-4-7 (deny-list)

### LINKED ISSUE #894: [BUG] OAuth Anthropic provider: clear_thinking_20251015 strategy forwarded without thinking block → 400 from upstream
## Summary

When Claude Code is configured with an OAuth-authenticated Anthropic provider in Tingly Box and the orchestrator model has extended thinking enabled (default for Opus 4.7), every request to `POST /tingly/claude_code/v1/messages?beta=true` fails with a 400 from upstream:

```
{"type":"error","error":{"type":"invalid_request_error","message":"`clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive"},"request_id":"req_011CasKQviFdqAv4S7D1CcqU"}
```

Anthropic is rejecting the upstream request because the body contains `context_management.strategy: clear_thinking_20251015` but no matching `thinking` block. Hypothesis: Tingly's protocol-translation pipeline preserves the `context_management` field while dropping the `thinking` field somewhere on the OAuth Anthropic path. The two must travel together.

The error is non-recoverable from the client side — Claude Code retries up to 10 times with exponential backoff, each retry burning quota against the user's Anthropic 7-day window before the session times out.

## Environment

- Tingly Box: **v0.260507.1** (binary at `~/.cache/tingly-box/v0.260507.1/bin/tingly-box`)
- Claude Code: **2.1.138**
- OS: Linux x86_64
- Provider auth type: **OAuth (Claude Code)** — Claude.ai Pro/Max subscription
- Provider name (auto-generated): `claude_code-YYYYMMDD-HHMM`
- Upstream model selected by routing: `claude-opus-4-7`
- Endpoint receiving traffic: `POST /tingly/claude_code/v1/messages?beta=true`

## Steps to reproduce

1. Add an OAuth Anthropic provider via the UI (`OAuth → Add Auth Provider → Claude Code`).
2. Run `tingly-box apply cc --provider <oauth-provider-name> --model claude-opus-4-7` (or wire `~/.claude/settings.json` manually) so Claude Code routes through `http://127.0.0.1:12580/tingly/claude_code` with `ANTHROPIC_DEFAULT_OPUS_MODEL=tingly/cc-opus`.
3. Configure the Tingly rule for `tingly/cc-opus` to point at the OAuth provider with model `claude-opus-4-7`.
4. Launch Claude Code with `"model": "opus"` in `~/.claude/settings.json` and the default thinking effort (High). Do **not** set `MAX_THINKING_TOKENS=0`.
5. Send any prompt.

Expected: a normal response from Anthropic Opus 4.7.

Actual: tight retry loop with 500s from Tingly wrapping the 400 above.

## Relevant log lines

From `~/.tingly-box/log/tingly-box.jsonl`:

```
[error] POST /tingly/claude_code/v1/messages?beta=true 500 — error: "Failed to forward request: POST \"https://api.anthropic.com/v1/messages?beta=true\": 400 Bad Request ... `clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive"
```

Earlier in the same log (after rule reconfiguration but before the strategy error surfaced), a related 500 also appeared with `CLAUDE_CODE_MAX_OUTPUT_TOKENS=32000`:

```
[error] POST /tingly/claude_code/v1/messages?beta=true 500 — error: "streaming is required for operations that may take longer than 10 minutes"
```

That one was sidestepped by lowering `CLAUDE_CODE_MAX_OUTPUT_TOKENS` from the apply-default of `32000` to `8192`. Worth noting because the apply CLI ships a default that Anthropic's policy now rejects without streaming — possibly a separate small bug.

## Workaround

Setting `MAX_THINKING_TOKENS=0` in `~/.claude/settings.json` env block disables Claude Code's extended thinking, which in turn stops Claude Code from attaching the `clear_thinking_20251015` strategy. Requests then succeed.

This is acceptable as a temporary workaround but disables Opus 4.7's reasoning quality on the orchestrator — the very feature most users pay for the subscription to get.

## Suggested fix

In the protocol-transform pipeline that handles the OAuth Anthropic path (likely `internal/protocol/transform/...` and the assembler/dispatch layer touched by recent PRs like #865, #875), preserve the `thinking` field alongside `context_management` when forwarding to upstream Anthropic. The two are coupled — if `context_management.strategy` is one of the `clear_thinking_*` family, `thinking` must be present (or removed together).

Either option resolves it:

- (a) Forward `thinking` alongside `context_management` whenever both are present in the incoming request.
- (b) Strip `context_management.strategy` matching `^clear_thinking_` whenever `thinking` is being stripped.

Option (a) preserves Claude Code behavior; (b) is safer if there's a reason `thinking` must be removed for OAuth.

## Related

- The `streaming is required` error suggests the apply CLI's default `CLAUDE_CODE_MAX_OUTPUT_TOKENS=32000` value is now too high for non-streaming paths. Consider lowering the default to `8192` in `internal/agent/apply.go::GenerateClaudeCodeEnv`.
- This blocks the canonical "Opus orchestrator + DeepSeek subagents" topology Tingly Box markets toward Claude Code users with a Pro/Max subscription.

Chee-0806:
## Summary

Fixes #894 — Claude Code OAuth provider sends `clear_thinking_20251015` in `context_management` alongside `thinking`, but the transform pipeline was stripping `thinking` for `claude-opus-4-7` (and any Gen 4.x model not explicitly listed) while preserving the `context_management` edits. Anthropic rejects this combination with a 400 error.

## Root cause

`isThinkingSupportedModel` used an **allow-list** that only matched `-4-6` suffix:

```go
strings.Contains(model, "claude-opus-4-6") || strings.Contains(model, "claude-sonnet-4-6")
```

`claude-opus-4-7` (and `claude-sonnet-4-7`, `claude-opus-4-5`, etc.) didn't match, so `thinking` was stripped but `context_management.clear_thinking_20251015` survived.

## Fix: deny-list on Gen 3

Changed to a **deny-list** that rejects `claude-3-*` (a closed, non-expanding model family) and passes everything else:

```go
func isThinkingSupportedModel(model string) bool {
    if !strings.HasPrefix(model, "claude-") {
        return false
    }
    return !strings.Contains(model, "claude-3-")
}
```

This is **future-proof** — Gen 4.x, Gen 5.x, Mythos, and any future model automatically get thinking support without code changes.

Also includes `stripBetaThinkingContextManagement` as a safety net.

### 1. CLAUDE_CODE_MAX_OUTPUT_TOKENS default (32000)

`internal/agent/apply.go` ships `CLAUDE_CODE_MAX_OUTPUT_TOKENS=32000` as default. On non-streaming paths, this triggers "streaming is required for operations that may take longer than 10 minutes". Should we lower the default to 8192?

### 2. Long-term: dedicated Claude Code client

As discussed in the issue, the OAuth round-trip is fragile. A dedicated Claude Code client (analogous to the Codex client) would eliminate the transform-pipeline dependency for OAuth paths altogether.

## Testing

- Unit tests for `isThinkingSupportedModel` cover Gen 3 deny, Gen 4+ pass, uppercase, date suffixes, empty/non-Claude edge cases
- `stripBetaThinkingContextManagement` unit tests
- Full transform/ops test suite passes
- No e2e testing with real Claude Code + OAuth Anthropic provider (requires credentials)

### DIFF
--- a/internal/protocol/transform/ops/request_anthropic_model.go
+++ b/internal/protocol/transform/ops/request_anthropic_model.go
@@ -17,8 +17,8 @@ const ClaudeCodeVersion = "2.1.86"
 const FingerprintSalt = "59cf53e54c78"
 
 // ApplyAnthropicV1ModelTransform applies Anthropic API v1 model-specific filtering.
-// This handles model-specific limitations such as adaptive thinking only being supported by
-// Claude Opus 4.6 (claude-opus-4-6) and Claude Sonnet 4.6 (claude-sonnet-4-6).
+// This handles model-specific limitations such as adaptive thinking not being supported by
+// Gen 3 models (claude-3-*). Gen 4+ models support adaptive thinking.
 //
 // Parameters:
 //   - req: The Anthropic v1 request to transform
@@ -36,7 +36,7 @@ func ApplyAnthropicV1ModelTransform(req *anthropic.MessageNewParams, model strin
 }
 
 // ApplyAnthropicBetaModelTransform applies Anthropic API beta model-specific filtering.
-// Same rules as V1 but for BetaMessageNewParams.
+// Same rules as V1 (Gen 3 deny-list) but for BetaMessageNewParams.
 func ApplyAnthropicBetaModelTransform(req *anthropic.BetaMessageNewParams, model string) *anthropic.BetaMessageNewParams {
 	if isThinkingSupportedModel(model) {
 		return req
@@ -45,10 +45,16 @@ func ApplyAnthropicBetaModelTransform(req *anthropic.BetaMessageNewParams, model
 }
 
 // isThinkingSupportedModel checks if the model supports adaptive thinking.
-// Only Claude Opus 4.6 and Claude Sonnet 4.6 support adaptive thinking.
+// Gen 3 models (claude-3-*) do NOT support it; Gen 4+ and future models do.
+// Uses a deny-list (negated match on "claude-3-") instead of an allow-list
+// because Gen 3 is closed and no new Gen 3 models will be released.
+// Non-Claude and empty models return false (safe default).
 func isThinkingSupportedModel(model string) bool {
 	modelLower := strings.ToLower(model)
-	return strings.Contains(modelLower, "claude-opus-4-6") || strings.Contains(modelLower, "claude-sonnet-4-6")
+	if !strings.HasPrefix(modelLower, "claude-") {
+		return false
+	}
+	return !strings.Contains(modelLower, "claude-3-")
 }
 
 // applyAnthropicV1ThinkingFilter removes thinking configuration from Anthropic v1 requests
@@ -75,8 +81,10 @@ func applyAnthropicV1ThinkingFilter(req *anthropic.MessageNewParams) *anthropic.
 	return req
 }
 
-// applyAnthropicBetaThinkingFilter removes thinking configuration from Anthropic v1 requests
-// // for models that don't support adaptive thinking.
+// applyAnthropicBetaThinkingFilter removes thinking configuration from Anthropic beta requests
+// for models that don't support adaptive thinking.
+// Also strips context_management.clear_thinking_20251015 edits that depend on thinking,
+// preventing 400 errors from Anthropic when thinking is removed but clear_thinking remains.
 func applyAnthropicBetaThinkingFilter(req *anthropic.BetaMessageNewParams) *anthropic.BetaMessageNewParams {
 	if req == nil {
 		return req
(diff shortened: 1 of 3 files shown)


internal/protocol/transform/ops/request_anthropic_model.go
```diff
@@ -96,6 +104,29 @@ func applyAnthropicBetaThinkingFilter(req *anthropic.BetaMessageNewParams) *anth
...
+	// Anthropic rejects the request with 400.
+	req = stripBetaThinkingContextManagement(req)
+
+	return req
+}
+
+// stripBetaThinkingContextManagement removes thinking-dependent context_management edits
+// from an Anthropic beta request. This prevents errors when thinking is stripped but
+// clear_thinking_20251015 edits remain, which Anthropic rejects as invalid.
+func stripBetaThinkingContextManagement(req *anthropic.BetaMessageNewParams) *anthropic.BetaMessageNewParams {
```
FFengIll, line 118: this may help.

internal/protocol/transform/ops/request_anthropic_model.go
```diff
@@ -45,10 +45,16 @@ func ApplyAnthropicBetaModelTransform(req *anthropic.BetaMessageNewParams, model
 }
 
 // isThinkingSupportedModel checks if the model supports adaptive thinking.
-// Only Claude Opus 4.6 and Claude Sonnet 4.6 support adaptive thinking.
+// Gen 3 models (claude-3-*) do NOT support it; Gen 4+ and future models do.
+// Uses a deny-list (negated match on "claude-3-") instead of an allow-list
+// because Gen 3 is closed and no new Gen 3 models will be released.
+// Non-Claude and empty models return false (safe default).
```
FFengIll, line 51: is there any reference (link) to confirm this?

0x0079:
I’ve checked these and made a low-level fix in #1075.

We don’t need to check the model itself; we only need to check the thinking config.

PR #1003, opened 2026-05-20
perf(network): Prevent env proxy inheritance when no explicit proxy configured

0x0079:
## Summary
This PR prevents HTTP clients from inheriting `HTTP_PROXY`/`HTTPS_PROXY` environment variables by default, improving security and predictability. The transport pool now respects an explicit `RespectEnvProxy` configuration flag (defaulting to `false`), and all client implementations (OpenAI, Anthropic, Google) are updated to use the transport pool instead of `http.DefaultTransport`.

## Key Changes

- **Transport Pool Configuration**: Added `RespectEnvProxy` field to `TransportConfig` to control whether environment proxy variables are inherited. Defaults to `false` for backward compatibility with secure defaults.

- **Dynamic Config Updates**: Modified `SetTransportConfig()` to detect changes to `RespectEnvProxy` and immediately clear cached transports so the new proxy policy takes effect on the next request.

- **Refactored Clear Logic**: Extracted `clearLocked()` helper method to allow both explicit `Clear()` calls and config-driven clearing to share the same logic while maintaining proper logging context.

- **Client Updates**: Updated `NewOpenAIClient()`, `NewAnthropicClient()`, and `NewGoogleClient()` to use `GetGlobalTransportPool().GetTransport()` instead of `http.DefaultTransport`, ensuring consistent proxy handling across all providers.

- **Startup Diagnostics**: Added `logProxyEnvironment()` method to log proxy-related environment variables and the `RespectEnvProxy` config value at startup and when config is applied, helping operators diagnose unexpected proxy usage.

- **Test Coverage**: Added three comprehensive tests:
  - `TestTransportPool_EnvProxyNotInherited`: Verifies env proxy is ignored by default
  - `TestTransportPool_EnvProxyRespectedWhenEnabled`: Verifies env proxy is used when `RespectEnvProxy=true`
  - `TestSetTransportConfig_ClearsPoolOnRespectEnvProxyChange`: Verifies cached transports are cleared when the setting changes

## Implementation Details

- The `boolPtrVal()` helper safely dereferences `*bool` pointers, treating `nil` as `false`
- Transport pool clearing distinguishes between immediately removed transports and those deferred (with active requests)
- All proxy environment variables (both uppercase and lowercase variants) are logged for comprehensive diagnostics

https://claude.ai/code/session_01At9Tny6A5bsQmCsPvJbJ4o

### DIFF
--- a/internal/client/anthropic.go
+++ b/internal/client/anthropic.go
@@ -79,6 +79,11 @@ func NewAnthropicClient(provider *typ.Provider, model string, sessionID typ.Sess
 			}
 			logrus.Infof("Using session-bound transport for OAuth issuer: %s, session: %s",
 				provider.OAuthDetail.GetIssuer(), sessionID.Value)
+		} else {
+			// OAuth provider with an issuer other than ClaudeCode (or missing OAuthDetail).
+			// Use a session-bound transport so proxy_url is respected and env proxy is
+			// not inherited — same guarantee as the non-OAuth path below.
+			transport = createSessionBoundTransport(provider, sessionID)
 		}
 	} else {
 		// Generic non-OAuth Anthropic provider. Apply the same User-Agent
--- a/internal/client/anthropic.go
+++ b/internal/client/anthropic.go
@@ -86,8 +86,12 @@ func NewAnthropicClient(provider *typ.Provider, model string, sessionID typ.Sess
 		// wraps innermost so it overwrites the header last, provider-UA sits
 		// outside. OAuth issuers above keep their dedicated transport chain
 		// unchanged because vendor-specific round-trippers manage UA themselves.
-		transport = http.DefaultTransport
-		transport = &customUserAgentTransport{base: transport}
+		//
+		// Use the transport pool instead of http.DefaultTransport so that env
+		// proxy variables (HTTP_PROXY / HTTPS_PROXY) are not inherited when no
+		// proxy is explicitly configured for the provider.
+		base := GetGlobalTransportPool().GetTransport(provider.UUID, model, provider.ProxyURL, ai.Issuer(""), sessionID)
+		transport = &customUserAgentTransport{base: base}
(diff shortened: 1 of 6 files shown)


internal/client/anthropic.go
```diff
@@ -86,8 +86,12 @@ func NewAnthropicClient(provider *typ.Provider, model string, sessionID typ.Sess
 		// wraps innermost so it overwrites the header last, provider-UA sits
 		// outside. OAuth issuers above keep their dedicated transport chain
 		// unchanged because vendor-specific round-trippers manage UA themselves.
-		transport = http.DefaultTransport
-		transport = &customUserAgentTransport{base: transport}
+		//
+		// Use the transport pool instead of http.DefaultTransport so that env
+		// proxy variables (HTTP_PROXY / HTTPS_PROXY) are not inherited when no
+		// proxy is explicitly configured for the provider.
+		base := GetGlobalTransportPool().GetTransport(provider.UUID, model, provider.ProxyURL, ai.Issuer(""), sessionID)
```
FFengIll, line 98: the global one get a general http transport in project.

internal/client/transport_pool.go
```diff
@@ -114,13 +114,30 @@ func GetGlobalTransportPool() *TransportPool {
...
+
 	globalTransportPool.config = config
 
+	newRespectEnvProxy := boolPtrVal(nil)
+	if config != nil {
+		newRespectEnvProxy = boolPtrVal(config.RespectEnvProxy)
+	}
+
+	if oldRespectEnvProxy != newRespectEnvProxy {
+		removed, deferred := globalTransportPool.clearLocked()
```
FFengIll, line 136: we need to refresh all transport cache to confirm the proxy active

FFengIll: a soft clear via lock 0

FFengIll (approved):

andreasfoo (approved):

PR #1070, opened 2026-05-29
Fix Claude Code local proxy bypass for Tingly endpoint

seviezhou:
## Summary

- add `NO_PROXY` / `no_proxy` defaults for Claude Code config generation
- ensure local Tingly endpoints (`localhost`, `127.0.0.1`, `::1`) bypass system HTTP proxies
- keep frontend Claude Code config preview aligned with backend env materialization
- cover no-proxy merging and default env behavior in tests

## Why

When `HTTP_PROXY` / `HTTPS_PROXY` are set globally, Claude Code can route `localhost:12580` requests through the proxy unless `NO_PROXY` is configured. In that state, requests never reach Tingly, so Tingly has no logs and Claude Code appears unable to connect to the model.

## Validation

- `GOCACHE=/private/tmp/tingly-go-cache go test ./internal/agent ./internal/command ./ai/agent ./internal/tbclient`
- `cd frontend && npx oxlint src/components/ClaudeCodeQuickConfig.tsx`

Note: `npm run typecheck` still fails on pre-existing unrelated repository type errors.

### DIFF
--- a/ai/agent/claude_code.go
+++ b/ai/agent/claude_code.go
@@ -59,6 +59,8 @@ func (p *ClaudeCodeParams) BuildEnv() map[string]string {
 		"API_TIMEOUT_MS":                           "3000000",
 		"ANTHROPIC_BASE_URL":                       p.BaseURL,
 		"ANTHROPIC_AUTH_TOKEN":                     p.APIKey,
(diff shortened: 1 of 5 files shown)


FFengIll:
This may have already been fixed in the main branch.

PR #1166, opened 2026-06-08
feat(clean_header+claude_code_compat): default-on for Claude Code & Desktop rules + OAuth suppression

0x0079:
## Summary

Defaults `clean_header` and `claude_code_compat` on for all built-in Claude Code **and** Claude Desktop rules, adds runtime suppression of `clean_header` for Claude OAuth providers, and removes `clean_header` from the scenario plugin surface entirely.

### `clean_header` — default-on for Claude Code & Desktop

Claude Code and Claude Desktop both inject `x-anthropic-billing-header` blocks into system messages. This is an Anthropic-internal billing mechanism; the header must never reach external providers.

Previously the header was only stripped during Anthropic→OpenAI protocol transformation (via `autoSetCleanHeaderFlag`). It is now default-on at rule level for both scenarios, so it applies unconditionally regardless of target protocol, and the Plugins card shows the true active state.

### `claude_code_compat` — default-on for Claude Desktop

Claude Desktop sends mid-conversation `role:"system"` messages (same non-standard extension as Claude Code) that third-party Anthropic-compatible providers reject. `claude_code_compat` was already default-on for Claude Code rules; this PR extends that to Claude Desktop.

### Claude OAuth suppression

When a rule routes to a Claude OAuth provider (native Anthropic subscription), `CleanHeader` is forcibly cleared at runtime in `resolveRuleFlagsWithScenario`. The billing header is consumed by Anthropic's billing backend for OAuth subscribers — stripping it would break subscription billing.

Uses the existing `provider.IsClaudeCodeProvider()` method (nil-safe).

### `clean_header` demoted from scenario plugin

`clean_header` is now rule-only. Removed `ScenarioFlags.CleanHeader` field, `GetScenarioFlag`/`SetScenarioFlag` branches, `FlagCleanHeader` constant, and the scenario OR-injection in `resolveRuleFlagsWithScenario`. The scenario-level flag had no remaining value once CC and Desktop rules default it on at rule level.

## Files changed

| File | Change |
|------|--------|
| `internal/server/config/init.go` | `ccRule()`: `ClaudeCodeCompat+CleanHeader: true`; new `cdRule()` factory with same flags; 4 inline Desktop structs replaced |
| `internal/server/config/config.go` | `newCCProfileRules`: `ClaudeCodeCompat+CleanHeader: true`; removed `GetScenarioFlag`/`SetScenarioFlag` branches for `clean_header` |
| `internal/server/config/migration.go` | `migrate20260608` (CC compat), `20260608_2` (CC clean_header), `20260608_3` (Desktop clean_header), `20260608_4` (Desktop compat); `migrate20260521` haiku rule updated |
| `internal/server/config/flag_keys.go` | Removed `FlagCleanHeader` constant |
| `internal/server/rule_flags.go` | `resolveRuleFlagsWithScenario` gains `provider` param; OAuth suppression added; scenario OR-injection removed |
| `internal/server/anthropic_message_v1.go` | Updated call site (pass `provider`) |
| `internal/server/anthropic_message_beta.go` | Updated call site (pass `provider`) |
| `internal/server/openai_chat.go` | Updated call site (pass `provider`) |
| `internal/server/openai_responses.go` | Updated call site (pass `provider`) |
| `internal/server/rule_flags_test.go` | Updated existing tests; new OAuth suppression test |
| `internal/typ/type.go` | Removed `ScenarioFlags.CleanHeader` field |
| `internal/typ/flag_registry.go` | `clean_header` spec: removed `Shared`/`InheritanceMode`; updated description |
| `.design/rule-flags.md` | Documented all changes |

## Notes

- `autoSetCleanHeaderFlag` is unchanged — it still auto-enables `CleanHeader` for `claude_desktop` during Anthropic→OpenAI protocol transformation.
- All four migrations are one-time gated (`hasMigrationCompleted`) so user overrides survive restarts.

### DIFF
--- a/internal/server/rule_flags.go
+++ b/internal/server/rule_flags.go
@@ -4,7 +4,6 @@ import (
 	"strings"
 
 	"github.com/gin-gonic/gin"
-	"github.com/tingly-dev/tingly-box/ai"
 	"github.com/tingly-dev/tingly-box/internal/protocol"
 	"github.com/tingly-dev/tingly-box/internal/protocol/transform"
 	"github.com/tingly-dev/tingly-box/internal/typ"
@@ -158,7 +157,7 @@ func resolveRuleFlagsWithScenario(
 	// consumed by Anthropic's billing backend; stripping it would break billing
 	// for OAuth subscribers even though it must be stripped for every other
 	// provider type (third-party Anthropic-compatible, OpenAI, etc.).
-	if flags.CleanHeader && isClaudeOAuthProvider(provider) {
+	if flags.CleanHeader && provider.IsClaudeCodeProvider() {
 		flags.CleanHeader = false
 	}
 
@@ -167,18 +166,6 @@ func resolveRuleFlagsWithScenario(
 	applyCustomUserAgent(c, flags)
 
 	return flags
-}
-
-// isClaudeOAuthProvider reports whether a provider routes to Anthropic via a
-// Claude Code OAuth token. These providers carry an x-anthropic-billing-header
-// that Anthropic's backend uses for subscription billing; it must not be stripped.
-func isClaudeOAuthProvider(provider *typ.Provider) bool {
-	if provider == nil {
-		return false
-	}
-	return provider.AuthType == typ.AuthTypeOAuth &&
-		provider.OAuthDetail != nil &&
-		provider.OAuthDetail.GetIssuer() == ai.IssuerClaudeCode
 }
 
 // applyCustomUserAgent attaches the effective custom User-Agent (already merged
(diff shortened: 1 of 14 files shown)


internal/server/rule_flags.go
```diff
@@ -149,6 +152,15 @@ func resolveRuleFlagsWithScenario(
 	// Auto-apply CleanHeader for protocol transformation in billing scenarios
 	flags = autoSetCleanHeaderFlag(flags, sourceAPI, targetAPI, scenarioType)
 
+	// Suppress CleanHeader when the provider is Claude OAuth (native Anthropic
+	// subscription). The x-anthropic-billing-header injected by Claude Code is
+	// consumed by Anthropic's billing backend; stripping it would break billing
+	// for OAuth subscribers even though it must be stripped for every other
+	// provider type (third-party Anthropic-compatible, OpenAI, etc.).
+	if flags.CleanHeader && provider.IsClaudeCodeProvider() {
```
0x0079, line 156: just provider.IsClaudeCodeProvider()?

PR #1181, opened 2026-06-09
fix(usage): preserve cache/reasoning tokens across protocol conversions

0x0079:
## Summary

Token usage was being lost on several protocol-**conversion** paths. The conversion code had richer usage (cache-read, cache-creation, reasoning) available, but narrowed it before it reached either the persisted analytics record or the client-facing response body. This PR threads the full usage through every wired conversion path.

The audit separated two axes:
- **Persisted analytics** — the `trackUsageWithTokenUsage` → DB/OTel record.
- **Response-body fidelity** — the `usage` block the client reads off the wire.

After this PR, no wired conversion path drops usage on either axis. (The persisted side was already correct everywhere *except* the streaming Anthropic→OpenAI path fixed below.)

### 1. Streaming Anthropic→OpenAI — persisted cache dropped to 0 (the original bug)
`AnthropicToOpenAIStream` / `…WithMCPHooks` returned `(int, int, error)`, so every caller rebuilt usage as `NewTokenUsageWithCache(in, out, 0)` — hardcoding cache to **0** in the persisted record, even though the internal `AnthropicAccumulator` had the cache-read count. The SSE response to the client was already correct; only the recorded analytics were wrong.
- Return `*protocol.TokenUsage` from both functions; record it directly at all three call sites (`protocol_dispatch.go`, `anthropic_to_openai_mcp_stream.go`). Error/cancel/EOF branches return the partially-accumulated usage so failures still record what was seen.

### 2. Nonstream Google→Anthropic — response body dropped cache
`ConvertGoogleToAnthropicResponse` / `…BetaResponse` built the usage block with only `input_tokens`/`output_tokens`, dropping Gemini's `CachedContentTokenCount`. Persisted analytics were already correct (extracted directly at dispatch); this is response fidelity — a Gemini-backed Anthropic client saw `cache_read_input_tokens=0`.
- Map `CachedContentTokenCount` → `cache_read_input_tokens` when present.

### 3. Nonstream Responses-API bodies — dropped detail / undercounted input
The nonstream Responses-API builders lagged their streaming counterparts:
- `BuildResponsesPayloadFromChat` omitted `input_tokens_details.cached_tokens` and `output_tokens_details.reasoning_tokens`.
- `BuildResponsesPayloadFromAnthropicBeta` reported `input_tokens` as the **uncached** slice only (undercount), and omitted `cached_tokens`. It now reports total prompt cost (`input + cache_read + cache_creation`) with `cached_tokens` detail — matching `toResponsesUsageWire` on the streaming path.

## Verification
- `go build` / `go vet` clean; `internal/protocol/stream`, `internal/protocol/nonstream`, `internal/protocol/usage`, and `internal/server` suites pass.
- Added regression tests: cache-read retained in the recorded usage (`TestAnthropicToOpenAIStream_VModelFullUsage`), Gemini cached-content → Anthropic cache-read, and both Responses-API builders carry cache/reasoning detail.

## Notes
Three Google converters (`ConvertAnthropicToGoogleResponse`, `ConvertGoogleToOpenAIResponse`, `ConvertOpenAIToGoogleResponse`) also drop cache but have **no dispatch caller** — latent dead code, left untouched here.

https://claude.ai/code/session_01X5bJk5NGbdyYWmwWgcqBtN

### DIFF
--- a/internal/protocol/nonstream/google_to.go
+++ b/internal/protocol/nonstream/google_to.go
@@ -172,10 +172,16 @@ func ConvertGoogleToAnthropicResponse(googleResp *genai.GenerateContentResponse,
 
 	// Add usage info if available
 	if googleResp.UsageMetadata != nil {
-		responseJSON["usage"] = map[string]interface{}{
+		usage := map[string]interface{}{
 			"input_tokens":  googleResp.UsageMetadata.PromptTokenCount,
 			"output_tokens": googleResp.UsageMetadata.CandidatesTokenCount,
 		}
+		// Gemini's cached-content tokens map to Anthropic cache-read so the
+		// client-facing response reports the cache hit instead of dropping it.
+		if googleResp.UsageMetadata.CachedContentTokenCount > 0 {
+			usage["cache_read_input_tokens"] = googleResp.UsageMetadata.CachedContentTokenCount
+		}
+		responseJSON["usage"] = usage
 	}
 
 	// Marshal and unmarshal to create proper Message struct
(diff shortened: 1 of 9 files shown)


FFengIll (approved):
LGTM

PR #1182, opened 2026-06-09
feat(rule): flexible per-rule 1M context for Claude Code & Codex

0x0079:
## What

Promotes the 1M (1,000,000-token) context window to a first-class **per-rule flag** (`context_1m`) that fans out from a single source of truth into three scenario-specific materializations.

Previously `[1m]` was only a decorative suffix the Claude Code Quick Config appended to a model string — the backend never understood it (rule matching is exact, so `tingly/cc-sonnet[1m]` matched no rule and 404'd), and there was no Codex equivalent.

Four parts:

1. **Per-rule 1M control** — a promoted **1M** switch on every rule card header (next to the request model), backed by the `context_1m` flag.
2. **Associated config** — the flag is the single source of truth; the Claude Code Quick Config derives the `[1m]` slot suffix from it; the generic snake↔camel flag conversion carries it with no per-field code.
3. **Per-agent materialization** — Claude Code injects the `context-1m-2025-08-07` anthropic-beta upstream (sent whenever requested, no model allow-list — let the upstream accept/reject); Codex widens the model-catalog `context_window` to 1,000,000 for flagged models.
4. **`[1m]` suffix preprocessing** — `typ.StripContext1MSuffix` strips the suffix before rule matching in both `determineRule` and `determineRuleWithScenario`, so messages and count-tokens resolve their rule.

## Toggle UX

Flipping 1M does not save silently — it opens a scenario-aware confirm dialog (`OneMConfirmDialog`) that spells out what takes effect where, with **one-click Apply** + **Cancel** (cancel reverts; the switch is controlled by the flag, written only on confirm).

Both CLI agents read their config at startup, so both get a symmetric **Apply + restart** flow:

- **Codex** → `Apply Codex config` (`api.applyCodexConfig()`, merged) → restart `codex` reminder.
- **Claude Code** → `Apply Claude Code config` (`applyClaudeCodeFromRules()` regenerates `settings.json` from the current rules via the same `derivePrefsFromRules` the Quick Config uses — same output as a default Quick Config apply, which already replaces the env block, so no new clobber) → restart Claude Code reminder.
- Other (anthropic/openai) scenarios are gateway-only → toggle directly, no dialog.

## How it's wired

- **Backend**: `RuleFlags.Context1M` + registry entry (persists/round-trips through the existing rule update path — no new API); `typ.WithContext1M`/`GetContext1M` (context-passing, mirroring `custom_user_agent`); `claudeRoundTripper` appends the beta when the request carries the intent; `RenderCodexModelCatalog(models, oneM)` widens the window, threaded through `ApplyCodexConfig` / `CodexParams` / `collectCodexRuleModels`.
- **Frontend**: `OneMContextSwitch`, `OneMConfirmDialog`, `RuleCard` state machine; `context_1m` hidden from the generic plugin catalog so it has one visual home. New UI strings use the project's paired zh/en local-map pattern.

Design notes: `.design/one-m-context.md`.

## Testing

- **Backend**: `go build ./...`, `go vet`, and package tests green. New tests cover the suffix helper, context propagation, the round-tripper beta injection, and the Codex catalog widening.
- **Frontend**: `tsc` + `oxlint` clean for changed files (repo baseline has unrelated codegen-artifact errors).
- **Visual**: headless-Chrome (mock mode) verified the 1M switch placement and both the Codex and Claude Code confirm dialogs.

https://claude.ai/code/session_01KZcRaXRvthMLWbW9CZGNyP


internal/agent/rule_bridge.go
```diff
@@ -193,6 +194,24 @@ func (aa *AgentApply) collectCodexRuleModels() []string {
 	return CollectCodexModels(models)
 }
 
+// collectCodexContext1MModels returns the set of active Codex-rule request
+// models whose context_1m flag is set. Keyed by request_model so it lines up
+// with the catalog slugs produced by collectCodexRuleModels / CollectCodexModels.
+func (aa *AgentApply) collectCodexContext1MModels() map[string]bool {
+	out := map[string]bool{}
+	for _, rule := range aa.config.GetRequestConfigs() {
+		if rule.GetScenario() != typ.ScenarioCodex || !rule.Active {
```
FFengIll: why special here for codex?

PR #1204, opened 2026-06-11
refactor(transform): make Vendor the final chain step; centralize assembly in BuildTransformChain

0x0079:
## What & why

The transform chain appended rule-driven "extra" transforms (`OpenAIMaxTokensRewriteTransform`, `RuleThinkingTransform`) to the **tail of the chain — after `Vendor` and after StagePost recording**. That broke two invariants:

- **Vendor must be the final, immutable step.** It directly faces the provider and does irreversible work (model alias, metadata, billing headers, DeepSeek thinking patch). Nothing should mutate the request after it.
- **Recording fidelity.** StagePost recording ran *before* those extras, so the recorded "final upstream request" didn't match what was actually dispatched.

## Changes

- **Centralize chain assembly in `BuildTransformChain`.** It now takes `preBase []transform.Transform` and `preVendor []transform.Transform` and slots them into fixed positions, with `Vendor` + recording as a guaranteed tail:
  ```
  preBase → StagePre-record → Base → MCP → Consistency → preVendor → Vendor → StagePost-record
  ```
  `preVendor` now runs after Consistency (so its param clamping still applies) and **before** Vendor — verified orthogonal to Vendor's work, so the move is behavior-preserving aside from the corrected ordering.

- **Remove per-handler chain mutation.** Dropped `prependPreBaseTransforms` / `appendExtraTransforms`; the four `transformXxxx` handlers (`transformAnthropicV1/Beta`, `transformOpenAIChat`, `transformOpenAIResponses`) just pass the two slices through. The builder is the single source of truth for ordering.

- **Rename + symmetry.** `ruleExtraTransforms` → `rulePreVendorTransforms`. The two dynamic insertion points are now symmetric — `preBase` (before Base) and `preVendor` (before Vendor) — both passed as plain slices (not variadic).

- **Tests.** New `internal/server/protocol_chain_builder_test.go` pins the canonical base order and guards that `preVendor` runs after `consistency_normalize` and strictly before `vendor_adjust`; uses stdlib `slices.Index`.

- **Docs.** `.design/rule-flags.md` updated with the slot model (`preBase` / `preVendor`), the canonical order, and the invariant: nothing runs after `Vendor` except recording.

## Notes / out of scope

`smart_compact` is still prepended by the two Anthropic handlers post-build (it applies to Anthropic inbound only); folding it into the builder would change behavior, so it's intentionally left as-is. Full `internal/server` and `internal/protocol/transform` suites pass.

https://claude.ai/code/session_01R9Gy8npezszdCGAGKGo5NK

### DIFF
--- a/.design/rule-flags.md
+++ b/.design/rule-flags.md
@@ -60,32 +60,35 @@
    │   flags := resolveRuleFlags(c, rule)                           │
    │   ├─ WithCustomUserAgent(ctx, flags.CustomUserAgent)           │  Type 2
    │   ├─ reqCtx.Extra["skip_usage"] = flags.SkipUsage              │  Type 3
-   │   ├─ preBase := rulePreBaseTransforms(flags)                   │  Type 1b-pre
+   │   ├─ preBase   := rulePreBaseTransforms(flags)                 │  Type 1b-pre
    │   │   → [transform.OpenAICursorCompatTransform{}]              │
-   │   └─ extras := ruleExtraTransforms(flags)                      │  Type 1b-post
+   │   └─ preVendor := rulePreVendorTransforms(flags)              │  Type 1b-post
    │       → [transform.OpenAIMaxTokensRewriteTransform{...}]       │
    └───────────────────────────┬───────────────────────────────────┘
                                │
                                ▼
    ┌───────────────────────────────────────────────────────────────┐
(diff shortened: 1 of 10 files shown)


FFengIll (approved):
LGTM

PR #1221, opened 2026-06-16
fix(server): eliminate access-log middleware OOM; bound in-memory diagnostics

0x0079:
## Problem

The access-log middleware buffered **every** response body in memory for the whole request lifetime — even for `200`s. Long-lived streaming LLM responses (the main proxy path) thus pinned their full size in the heap; with concurrency this scaled to multi-GB OOM over days.

Measured pre-fix (`87786d3`): a single 64 MiB stream pinned **63 MiB**, and concurrent in-flight streams scaled **linearly (~16 MiB × N, 255 MiB at N=16)**.

## Root fix — status-gated, lazily-allocated response capture

- `responseBodyWriter` writes through first (so `Status()` is known), then captures **only when `status >= gate`** (default 400). The buffer is **never allocated on the 200 happy path**. No stream detection — purely status-driven.
- The gate drops to 0 only when an admin bad-request filter targets sub-400 statuses (`CapturesBelow400`, cached per filter change), still bounded by the cap.

After fix: 200 streams retain **0 MiB** regardless of size or concurrency.

## Supporting changes

- **Consolidation:** merge the duplicated error-log + access-log middlewares into one (delete `error_log.go`), extracting a dedicated `BadRequestSink` that rotates via the standard **lumberjack** rotator (no bespoke stat/rename/glob logic). *Net −90 production lines.*
- **Pooling + smart defaults:** pool request-mirror buffers; capture caps configurable with smart defaults (`0` = built-in), not user-facing knobs.
- **Memory bounded by bytes, not just count:** total-byte budget with LRU eviction (keep ≥1) on `MemoryLogHook` and `RequestBodyStore`; a single body is truncated only as a last resort.
- **Disk fidelity decoupled from memory bounds:** the bad-request sink records the full captured body while the in-memory store may truncate; truncation flagged on both (`request_body_truncated` / `response_body_truncated`).

## Tests

- **`BuffersEntireStreamingResponse`** — living canary: self-skips on fixed code, reproduces (~64 MiB pinned) if the gate is reverted.
- **`StreamingResponseNotBuffered` / `ConcurrentStreamsStayFlat`** — 200 streams retain ~0 regardless of size/concurrency.
- **`ErrorResponseCaptureBounded`** — 5xx bodies still captured but clipped at the cap, with the truncation flag.
- Byte-budget eviction + last-resort truncation tests for both stores; decoupled disk-full/memory-truncated fidelity test.

## Verification

`go build ./...`, `go vet`, and `go test ./pkg/obs/... ./internal/server/middleware/... ./internal/server/ -race` all green.

| | Pre-fix `87786d3` | This PR |
|---|---:|---:|
| Single 64 MiB stream retained | 63 MiB | **0 MiB** |
| 16 concurrent 16 MiB streams | 255 MiB | **0 MiB** |
| 8 MiB 5xx body captured | unbounded | 64 KiB (truncation flagged) |

https://claude.ai/code/session_01DqgVpbJvEXDF5FQzojTPvY

---
_Generated by [Claude Code](https://claude.ai/code/session_01DqgVpbJvEXDF5FQzojTPvY)_


internal/server/config/config.go
```diff
@@ -72,6 +72,10 @@ type Config struct {
 	// Error log settings
 	ErrorLogFilterExpression string `json:"error_log_filter_expression"` // Expression for filtering error log entries (default: "StatusCode >= 400 && (Path matches '^/api/' || Path matches '^/tbe/')")
 
+	// HTTPLogCapture controls in-memory diagnostic capture of request/response
+	// bodies by the access-log middleware. Zero values use built-in defaults.
+	HTTPLogCapture HTTPLogCaptureConfig `json:"http_log_capture,omitempty" yaml:"http_log_capture,omitempty"`
```
0x0079: rename: HTTPCaptureXXX prefix is enough, non log literal

internal/server/middleware/bad_request_sink.go
```diff
@@ -0,0 +1,273 @@
+package middleware
```
0x0079, line 1: does this file overdesign ?

internal/server/middleware/multi_mode_memory_log.go
```diff
@@ -15,14 +15,58 @@ import (
...
 )
 
+// captureNever is a sentinel response-capture gate that no real HTTP status can
+// reach, used to disable response-body capture entirely.
+const captureNever = 1 << 30
+
+// CaptureConfig tunes the middleware's in-memory body capture. Zero size fields
+// fall back to the package defaults; Disabled turns body capture off entirely
+// (request metadata + status code are still logged).
+type CaptureConfig struct {
```
0x0079, line 40: if we can limit request count and total store, should we limit body size again? judge first

internal/server/server.go
```diff
@@ -58,7 +58,7 @@ type Server struct {
 	multiLogger *pkgobs.MultiLogger
 
 	// middleware
-	errorMW         *middleware.ErrorLogMiddleware
+	badReqSink      *middleware.BadRequestSink
```
0x0079: I belive this sink belong to mw, not server.
server just config and init mw for all of these

internal/server/config/config.go
0x0079: Done — renamed to `HTTPCaptureConfig` (field `HTTPCapture`, tag `http_capture`, getter `GetHTTPCapture`). Safe since the field is new in this PR. (5f22ac7)

---

internal/server/server.go
0x0079: Agreed — moved ownership to the mw. Removed `Server.badReqSink`; the middleware now owns the sink and exposes `SetBadRequestFilter` + `Close`. Server only builds/configures the mw and routes filter-reload + shutdown through it. (5f22ac7)

---

internal/server/middleware/bad_request_sink.go
0x0079: Judged & trimmed. The bulk here is **carried-over functionality**, not new design: the expr filter (`FilterContext` / `SetFilterExpression` / `shouldLog`) is the existing configurable bad-request filter moved out of the old `ErrorLogMiddleware`, and rotation is now delegated to `lumberjack` (the bespoke stat/rename/glob logic was already deleted).

Concrete simplification this round: dropped the dead json tags on `logEntry` — it's an internal DTO; `maybeLog` is the single serializer (bodies as raw JSON when valid, else quoted; duration in ms). (5f22ac7)

The one genuinely "clever" bit is `computeMatchesSub400` (probing the compiled filter to decide whether to also capture <400 response bodies). It's small and keeps the 200 happy path allocation-free for the default filter, but it only matters if an admin writes a sub-400 filter. If you'd rather drop that capability entirely (custom sub-400 filters would then log without response bodies), I'm happy to remove it — say the word.

---

internal/server/middleware/multi_mode_memory_log.go
0x0079: Judged — they bound different things, so I kept it:

- `MaxRequestBodySize` is the **capture cap**: how much of a single body we mirror **in-flight** during the request. It bounds transient memory (≈ concurrency × cap) and is the fidelity ceiling for the on-disk sink. Without it, one large upload buffers in full before anything truncates it.
- `MaxRequestBodies` (count) + the store byte budget bound **retention** — what's kept across requests after the fact.

So it's "per-body in-flight cap" vs "total retained" — not the same limit. The store truncates a body only as a last resort (when one body alone exceeds the whole budget).

If the goal is fewer knobs, the better candidate to drop is the store **count** cap (`MaxRequestBodies`), since the byte budget already bounds retention — count is just cheap insurance. Want me to drop it?

---

internal/server/middleware/bad_request_sink.go
0x0079: re-think about the full file, the filter seems overdesign ?

0x0079: Agreed — removed it. Since the filter was never actually adjusted in practice (config-file only, no UI), the expr engine was pure overdesign. Replaced with a fixed predicate `status >= 400 && (path /api|/tbe)`:

- `bad_request_sink.go`: hardcoded `shouldLog`; dropped `FilterContext`, `expr.Compile/Run`, `SetFilterExpression`, `computeMatchesSub400` / `CapturesBelow400`, and the RWMutex.
- middleware: removed `SetBadRequestFilter` and the sub-400 capture-gate branch — the gate is now always 400, 200 happy path stays allocation-free.
- `server.go`: dropped the filter init + config-reload wiring.
- `config.go`: removed the dead `ErrorLogFilterExpression` field/getter/setter/defaults.
- `go.mod`: dropped the now-unused `expr-lang` dependency.

The file is now just: open a lumberjack file → predicate → write one JSON line. (7c40842)

---

PR #1298, opened 2026-07-03
refactor(server): extract modules and separate protocol and web handlers which help reduce files in server.

FFengIll:
We mix both protocol handler and web handler into server methods, it is hard to manage and extend (more than 80 files).

This PR do some refactoring and separate them into protocol and web handler structure with deps definition.

It can help
- Manage files and provide a foundation for future splitting
- Facilitate independent testing (instead of heavily relying on the server)
- Clarify dependencies and lay the foundation for future expansion
- Code clean to reduce noise


0x0079:
@codex check consistent

Issue #1316, opened 2026-07-06
[BUG] CC API Error: API returned an empty or malformed response (HTTP 200)


kz25-1766286:
### Version

0.260709.0-rc1

### OS & Architecture

Linux x86_64

### Category

Protocol

### Scenario

I was using claude code via tingly box codex oauth. Randomly showing
`API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request`
Continue working after entering retry to CC.

### Logs / Error Output

```shell

```

hiSandog:
Because retry succeeds, the diagnostic should capture the response's content type, content length, upstream request id, and whether the body was empty versus invalid JSON—without logging tokens or response content. That would distinguish a gateway-generated 200 from a truncated streaming response. Retrying once is reasonable only before any assistant content has been delivered; after partial output, the client should avoid silently duplicating a request.

FFengIll:
In #1317, we have addressed the issue where an empty stream was returned when an error occurred. 

While we are unable to confirm this as the root cause at this time, we welcome further feedback to help validate the fix

kz25-1766286:
0.260716.0-rc3 still exists
`API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request`

FFengIll:
For claude code, it will call nonstream in some edge cases (stream: false), but codex do not support nonstream.
So we have to do stream request and assemble them into valide nonstream responses.

#1381 and #1378 introduced why it raise malformed response error, and fixed this corresponding problem,

PR #1378, opened 2026-07-17
bugfix(protocol): Fix assembly response handling to prevent malformed JSON errors

andreasfoo:
## Summary
Fixes issue #1316 where the Responses API to Anthropic assembly handler was returning responses with `text/event-stream` Content-Type headers, causing strict SDK clients (like Claude Code's non-streaming fallback) to skip JSON parsing and treat the body as opaque text, resulting in "API returned an empty or malformed response (HTTP 200)" errors.

## Key Changes

- **Removed `text/event-stream` Content-Type header** from assembly responses. The assembled response is a single JSON message, so gin's JSON render now correctly sets `application/json`, allowing SDK clients to parse the response properly.

- **Changed empty content handling** to fail with HTTP 502 (Bad Gateway) instead of returning HTTP 200 with `content: null`. This prevents strict clients from rejecting the malformed message and allows the failover orchestrator to retry another candidate.

- **Added `failEmptyAssembly()` function** to report upstream streams that produced no usable content blocks with appropriate error details and retryable status codes.

- **Added `normalizeAssemblyStopReason()` function** to fill in a stop reason for streams cut before their terminal event: `tool_use` if a tool call completed, `end_turn` otherwise. This ensures assembled messages are well-formed even when truncated after partial content.

- **Updated panic recovery** in `recoverAssemblyPanic()` to return a proper JSON error response instead of an SSE error frame, and only respond if headers haven't been written yet.

- **Simplified assembly headers** by removing `Connection: keep-alive` (not needed for non-streaming responses) and the problematic `text/event-stream` Content-Type.

- **Updated test cases** to verify correct Content-Type headers and distinguish between two truncation scenarios:
  - `TruncatedNoContent`: Stream cut before any content block completes → HTTP 502 error
  - `TruncatedPartialContent`: Stream cut after content blocks completed → HTTP 200 with normalized stop_reason

## Implementation Details

The fix addresses a fundamental mismatch: the assembly path produces a single JSON response, not a stream of SSE events. The historical `text/event-stream` header was a copy-paste artifact from the streaming implementation that broke strict JSON parsing in downstream clients. By removing it and properly handling edge cases (empty content, truncation), the response now works correctly with all client types while maintaining proper error semantics for failover retry logic.

### DIFF
--- a/internal/protocol/stream/openai_responses_to_anthropic_assembly.go
+++ b/internal/protocol/stream/openai_responses_to_anthropic_assembly.go
@@ -2,6 +2,7 @@ package stream
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/anthropics/anthropic-sdk-go"
@@ -123,15 +124,14 @@ func (a *responsesToAnthropicAssembly) run(c *gin.Context, stream ResponsesStrea
 	return conv.Usage(), stream.Err()
 }
 
-// setAssemblyHeadersAndRecover mirrors the header and panic behavior of the
-// previous assembly implementation: SSE headers are set up-front (gin's JSON
-// render preserves an already-set Content-Type, so responses keep the
-// historical text/event-stream header), and a panic surfaces as an SSE error
-// frame.
+// setAssemblyHeaders sets the response headers for the assembly path. The
+// assembled response is a single JSON message, so the Content-Type is left to
+// gin's JSON render (application/json). A historical implementation forced
+// text/event-stream here, which made strict SDK clients (e.g. Claude Code's
+// non-streaming fallback) skip JSON parsing and treat the body as opaque text,
+// surfacing as "API returned an empty or malformed response (HTTP 200)" (#1316).
 func setAssemblyHeaders(c *gin.Context) {
-	c.Header("Content-Type", "text/event-stream")
 	c.Header("Cache-Control", "no-cache")
-	c.Header("Connection", "keep-alive")
 	c.Header("Access-Control-Allow-Origin", "*")
 	c.Header("Access-Control-Allow-Headers", "Cache-Control")
 }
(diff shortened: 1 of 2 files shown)


internal/protocol/stream/openai_responses_to_anthropic_assembly.go
```diff
@@ -123,29 +124,63 @@ func (a *responsesToAnthropicAssembly) run(c *gin.Context, stream ResponsesStrea
...
+			Type:    "api_error",
+		},
+	})
+	return err
+}
+
+// normalizeAssemblyStopReason fills in a stop reason for a stream that was cut
+// before its terminal Responses event: tool_use if a tool call completed,
+// end_turn otherwise. Only called when at least one content block assembled.
+func normalizeAssemblyStopReason(asm *responsesToAnthropicAssembly) string {
```
0x0079, line 174: a guard here?
does it happen?

internal/protocol/stream/openai_responses_to_anthropic_assembly.go
```diff
@@ -123,29 +124,63 @@ func (a *responsesToAnthropicAssembly) run(c *gin.Context, stream ResponsesStrea
...
-// historical text/event-stream header), and a panic surfaces as an SSE error
-// frame.
+// setAssemblyHeaders sets the response headers for the assembly path. The
+// assembled response is a single JSON message, so the Content-Type is left to
+// gin's JSON render (application/json). A historical implementation forced
+// text/event-stream here, which made strict SDK clients (e.g. Claude Code's
+// non-streaming fallback) skip JSON parsing and treat the body as opaque text,
+// surfacing as "API returned an empty or malformed response (HTTP 200)" (#1316).
 func setAssemblyHeaders(c *gin.Context) {
-	c.Header("Content-Type", "text/event-stream")
```
0x0079, line 132: this might be the root cuase.

0x0079 (approved):

PR #1433, opened 2026-07-25
refactor(imbot): platform seams — neutral outbound actions, capability table, message restating

0x0079:
Extracting imbot flattened the **connection** layer's platform differences but not the **interaction** layer, so `internal/remote_control` was still acting as a renderer and a capability judge. This lands phases 1, 2a and 3 of the seam work recorded in `.design/imbot-platform-seams.md`.

## Root cause

> imbot abstracts **inbound** messages (`core.Message` normalisation) and the **send action** (`SendMessage`), but not the **outbound payload**. That travelled through an untyped `Metadata map[string]interface{}`, so every caller had to know which platform it was talking to.

The historical shape — three renderings of the same buttons at one call site, hoping the right one got picked up:

```go
metadata["replyMarkup"] = tgKeyboard                     // Telegram shape
metadata["card"]        = card                           // neutral shape
if platform == Feishu { metadata["card_json"] = ... }    // Feishu shape
```

The last two were never read on any send path. The first carried a go-telegram type Feishu could not decode.

## Defects fixed

**Inline buttons were invisible on Feishu/Lark.** `remote_control` always sent a `models.InlineKeyboardMarkup`; Feishu's type switch accepted only `interaction.InlineKeyboardMarkup` and `map[string]interface{}`, matched neither, and produced a card with **no buttons at all**. Clear / CD / Project, the directory browser and the `/resume` picker were all unreachable for Feishu users. The 161-line Feishu renderer whose output fed `card_json` was dead code.

**Lark bots could not start.** Both hand-written switches in the bot manager omitted `"lark"` while `PlatformConfigs` had an entry for it. Two independent failures from one cause: the credential check fell to the default branch and demanded a token the Lark form never collects, and `buildAuthConfig` would have handed the Feishu client a token-type config it rejects.

**Keyboards were never removed off Telegram.** `AsTelegramBot` was a *concrete* type assertion (`bot.(*telegram.Bot)`), so Feishu and tingly could never satisfy it — users could re-click a used menu into stale state, which the code's own comment said the removal existed to prevent.

**The verbose capability check was commented out** with its doc comment left behind, so what the code did and what it claimed had drifted.

**The Weixin QR client was duplicated** — `imbot/platform/weixin.QRClient` (used by the web module, tested) and `feature.WeChatQRClient` (used by the CLI, untested) — and the two had diverged on timeout handling.

## Design

**Three tiers, so the neutral model does not swallow what should stay platform-specific:**

| Tier | Example | Treatment |
|---|---|---|
| Universal | clickable action, open link | abstracted — `core.Action`'s own fields |
| Capability-gated | open a mini app | neutral semantics, platform picks the implementation, `Fallback` is explicit |
| Platform-only | `pay`, `switch_inline_query` | **not** abstracted — typed escape hatch via `Action.Ext` and constructors in the platform package |

Tier 3 shipped in the same change as the removal of `Metadata["replyMarkup"]`, because that was the only existing route to a Telegram-specific button; closing it without opening another would have been a capability regression. The point was never to eliminate platform-specific code, only the *invisible* kind: `telegram.WebAppButton(...)` declares itself through its import and stays greppable.

**Seams landed:**

- **Actions** — `SendMessageOptions.Actions *core.ActionSet`, rendered per platform. Rows are preserved because layout is meaningful.
- **Capability table** — `PlatformDescriptor.Behavior` and `PlatformAuthConfig.Auth` replace six switch statements. `AuthMapping` is deliberately *not* derived from `FieldSpec`: `FieldSpec` describes the settings-UI form while this is the wire mapping, and Weixin's credentials come from the QR flow with no form fields at all.
- **Restating** — `core.MessageRestater` expresses intent ("this message and its buttons are stale") rather than mechanism. Telegram edits in place; Feishu patches the card, which updates what the user already sees without a new notification.

## Known limits

- Feishu's card patch replaces the whole content, so it cannot drop buttons while leaving the body alone. It reports `ErrNotSupported` for that rather than blanking the user's message, which makes the keyboard-removal fix partial there — the three pure "take the menu down" sites stay no-ops on Feishu.
- **Weixin no longer sends intermediate progress messages.** Each outbound message is tied to one inbound reply context, so a running commentary cannot be delivered. This restores the documented intent rather than introducing a new policy.
- The web QR flow is now slightly more tolerant of a timed-out long poll, taking the more robust of the two deduplicated implementations.
- The Feishu button fix is derived from the type switch and is deterministic, but has not been confirmed against a live Feishu bot — no credentials in this environment.

## Not in this change

`Action.Payload` replacing `CallbackData` is the next step and touches the callback protocol, so it belongs in its own PR. Telegram's 64-byte `callback_data` limit has already leaked into the application architecture — the directory browser navigates by index and snapshots `BindFlowState.Dirs` purely to fit it, a constraint Feishu does not have — and the limit itself is documented in a comment but never validated, so `bind:create:<path>` fails the whole send for paths over 52 bytes.

`SendMessageOptions.Card` and per-action `Style` are deferred: the Card type family has 48 external references and no send path consumes it yet. Reply-context forwarding, the `menu` package, and removal of the deprecated metadata path follow in phases 4–6; the compat path cannot go first, since imbot's own interaction handler and the `menu` adapters still produce `replyMarkup`.

## Verification

Build and `go vet` clean with and without `-tags e2e` — worth noting that `imbot/tests/` sits behind the `e2e` build tag, so a plain `go test ./...` does not compile it and can look green while the tree is broken. 92 packages pass, 0 failures.

New tests cover the Feishu button regression, the Lark auth mapping, Tier 3 fallback behaviour on a foreign platform, restating on a non-Telegram platform, and a guard that fails if a future platform is added to the config table without a wire mapping.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QbXSY3DYEhpre9hrVQagkt)_


0x0079:
Superseded by a three-PR stack — this branch was too large to review as one unit (46 files, +2334).

- #1434 — 1/3: ownership moves
- #1435 — 2/3: neutral outbound actions
- #1436 — 3/3: capability table and message restating

Same final tree, with two differences: the stack is rebased onto current `main`, and it drops gofmt churn in three files (`pairing_manager.go`, `chat_store_pair_test.go`, `tingly_e2e_test.go`) that were already unformatted on `main` and that this work never touched.

Closing in favour of the stack.

---

PR #1445, opened 2026-07-26
fix(protocol): use fc-prefixed item IDs for Responses function calls

E-lai:
## Summary
Chat Completions exposes one tool-call ID (`call_...`), while Responses uses separate fields for the function-call item ID (`id`, `fc_...`) and the tool-call correlation ID (`call_id`). The converter replaced the generated `fc_...` item ID with `call_...`, so native Responses endpoints rejected the item when conversation history was replayed.

This change keeps `fc_...` as the item `id`, preserves `call_...` as `call_id`, and adds regression coverage for argument-event `item_id` references.

## Log

```
base_url="https://chatgpt.com/backend-api", returning HTTP 400: Invalid 'input[67].id': 'call_...'. Expected an ID that begins with 'fc'.
```

## Tests

```text
go test ./internal/protocol/stream -count=1
go test ./internal/protocol/request ./internal/protocol/nonstream -count=1
```

### DIFF
--- a/internal/protocol/stream/openai_chat_to_responses_converter.go
+++ b/internal/protocol/stream/openai_chat_to_responses_converter.go
@@ -143,10 +143,15 @@ func (c *chatToResponsesConverter) processChunk(chunk *openai.ChatCompletionChun
 		openaiIndex := int(toolCall.Index)
 
 		if _, exists := c.pendingToolCalls[openaiIndex]; !exists {
+			// A Responses function-call item has two distinct identifiers:
+			//   - id: the Responses output item ID (must use the fc_ prefix)
+			//   - call_id: the correlation ID used by function_call_output
+			//
+			// Chat Completions exposes only tool_call.id (normally call_...).
+			// Preserve it as call_id, but never reuse it as the Responses item ID:
+			// Codex persists output item IDs and a later native Responses request
+			// rejects an item whose id starts with call_ instead of fc_.
 			itemID := fmt.Sprintf("fc_%d_%d", time.Now().Unix(), openaiIndex)
-			if toolCall.ID != "" {
-				itemID = truncateToolCallID(toolCall.ID)
-			}
 
 			toolOutputIndex := c.outputIndex
 			c.outputIndex++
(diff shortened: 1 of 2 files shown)


FFengIll:
@andreasfoo

FFengIll:
@E-lai thanks for your PR, and it makes sense.

Furthermore, can you determine the error source - by client or by server?

```
Invalid 'input[n].id': 'call_123'. Expected an ID that begins with 'fc'.
```

E-lai:
Thanks for asking. I traced the request end to end, and the answer is layered:

- **The validation error itself is returned by the upstream server.** In a real failure on August 10, the Tingly Box guardian log records the request at `stage=upstream`, with `base_url="https://chatgpt.com/backend-api"`, returning HTTP 400: `Invalid 'input[67].id': 'call_...'. Expected an ID that begins with 'fc'.`
- **The malformed item is produced by Tingly Box's Chat Completions -> Responses conversion.** The converter was using the Chat `tool_call.id` (`call_...`) as both the Responses item `id` and the correlation `call_id`.
- **The client only persists and replays that converted output item.** On a subsequent turn, Codex sends the stored item back; a native Responses endpoint then validates `input[n].id` and rejects it.

So, in short: **the server emits the error, but the protocol-conversion bug in Tingly Box creates the invalid input**. The client makes the issue persistent/visible by replaying conversation history, but it does not originate the invalid ID.

The intended mapping is:

```text
Chat tool_call.id (call_...) -> Responses function_call.call_id
Generated Responses item ID (fc_...) -> Responses function_call.id and event item_id
```

I have also rebased this PR onto the latest `main` (`85f1126`) and resolved the golden-test event-index conflict introduced by the newer stream-converter event sequence. The relevant tests pass on the rebased branch:

```text
go test ./internal/protocol/stream -count=1
go test ./internal/protocol/request ./internal/protocol/nonstream -count=1
```

FFengIll (approved):

FFengIll:
Thanks for your PR.

BTW, before merge, please update your PR content to a concise version if possible.  @E-lai

E-lai:
Updated, thanks. I condensed the PR description to the essential problem, fix, and test commands, while leaving the detailed tracing in the discussion for reference.

FFengIll:
> Updated, thanks. I condensed the PR description to the essential problem, fix, and test commands, while leaving the detailed tracing in the discussion for reference.

Thanks for the follow-up, and I add the log into description too for tracking.
We are going to merge.

PR #1462, opened 2026-07-31
fix(oauth): forward anthropic-organization-id on Claude Code OAuth requests

andreasfoo:
## Problem

When a Claude Code OAuth provider is used, requests forwarded to Anthropic never carry the organization identity, so Anthropic attributes them to the token's **default/personal context** instead of the organization the token was issued for. Any entitlement bound to a specific organization (e.g. the Cyber Verification Program) silently stops applying, and requests get rejected by org-level safeguards even though the account is verified.

Root cause: the org id was **write-only**.

- At OAuth login, `populateAnthropicMetadata` (`ai/oauth/manager.go`) extracts `organization_id` from Anthropic's token response and persists it into `OAuthDetail.ExtraFields`.
- At request time, `claudeRoundTripper.applyClaudeCodeHeaders` (`internal/client/claude_round_tripper.go`) deletes **every** `anthropic-*` header and rebuilds a fixed header table — which did not include `anthropic-organization-id`. Nothing in the codebase ever read the stored `organization_id` back.

So the value was captured and stored on login, then dropped on every outgoing request.

## Fix

Read `organization_id` from `OAuthDetail.ExtraFields` and attach it as the `anthropic-organization-id` header on both Claude Code OAuth request paths:

1. **`claudeRoundTripper` path** (`internal/client/anthropic.go` + `claude_round_tripper.go`): the round tripper gains an `organizationID` field, populated at construction from the provider. It is added to the rebuilt header table inside `applyClaudeCodeHeaders` — it *must* live there rather than in SDK options, because the header cleanup strips all `anthropic-*` headers before the table is applied.
2. **`ClaudeClient` SDK-options path** (`internal/client/claude_client.go`): the package-level `applyClaudeCodeHeaders` appends the same header via `anthropicOption.WithHeader`.

The header is only set when a non-empty `organization_id` exists, so:
- API-key providers and non-ClaudeCode OAuth issuers are untouched.
- Older stored OAuth providers without the metadata behave exactly as before.
- A stale upstream-supplied `anthropic-organization-id` is still stripped by the cleanup and not resurrected.

This mirrors the existing precedent in `codex_client.go`, which forwards `account_id` from ExtraFields as `X-ChatGPT-Account-ID`.

## Testing

- New unit test `TestApplyClaudeCodeHeadersOrganizationID` covers both cases: the header is forwarded when the org id is present, and absent (including stripping a stale upstream value) when it is not.
- `go test ./internal/client/` passes.

### DIFF
--- a/internal/client/anthropic.go
+++ b/internal/client/anthropic.go
@@ -76,7 +76,8 @@ func NewAnthropicClient(provider *typ.Provider, model string, sessionID typ.Sess
 	if provider.AuthType == typ.AuthTypeOAuth {
 		if provider.OAuthDetail != nil && provider.OAuthDetail.Issuer == ai.IssuerClaudeCode {
 			transport = &claudeRoundTripper{
-				RoundTripper: createSessionBoundTransport(provider, sessionID),
+				RoundTripper:   createSessionBoundTransport(provider, sessionID),
+				organizationID: provider.OAuthDetail.GetExtraFieldString("organization_id"),
 			}
 			logrus.Infof("Using session-bound transport for OAuth issuer: %s, session: %s",
 				provider.OAuthDetail.GetIssuer(), sessionID.Value)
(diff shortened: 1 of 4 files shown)


0x0079 (approved):
LGTM

PR #1466, opened 2026-08-01
feat(oauth): claude_org_id rule flag for per-rule organization control

0x0079:
## Summary

Follow-up to #1462, which started forwarding the OAuth login organization as `anthropic-organization-id` on every Claude OAuth request — a behavior change for all existing setups, and the value is immutable after login. This makes organization attribution an explicit, per-rule choice via a new `claude_org_id` rule flag:

- **unset (empty, default)** — send no organization header, restoring the classic pre-#1462 behavior
- **`auto`** — send the organization captured at OAuth login (needed for org-bound entitlements like Cyber Verification)
- **any other value** — send that organization id verbatim

## Key Changes

- `typ.RuleFlags.ClaudeOrgID` + registry entry (`request_anthropic` category, `Login organization` quick-pick via the `typ.ClaudeOrgIDAuto` sentinel) — registry-driven catalog, no dedicated frontend work needed.
- `ResolveRuleFlagsWithScenario` attaches the flag to the request context, following the existing `context_1m` / `custom_user_agent` hint pattern.
- `NewClaudeClient` resolves the flag when composing `anthropic-organization-id` (constructed per request, so it's correctly request-scoped). Generic Anthropic path is untouched — this is Claude OAuth specific.
- `claudeOrgId` / `claude_org_id` typings in `RoutingGraphTypes.ts`.

## Compatibility Notes

- **Default behavior changes vs #1462**: with the flag unset, Claude OAuth requests no longer carry `anthropic-organization-id`. Rules that need login-org attribution must set `auto`.

## Testing

- Gateway-level `claude_org_id` case on the real Claude OAuth path (pool → `NewClaudeClient`) pins all three states against the captured upstream request.
- `TestClaudeClient_ClaudeOrgID` drives `NewClaudeClient → Guard → SDK → HTTP` against a capturing server across the same matrix.
- `go test ./internal/typ/... ./internal/client/... ./internal/server/ ./internal/protocoltest/` and `go vet ./internal/...` pass.

## Notes

- `openapi.json` intentionally not regenerated — a regen currently pulls in ~500 lines of unrelated drift and breaks the frontend quota typecheck. Needs a separate `task codegen` pass; the flag works without it since rule flags round-trip through `typ.RuleFlags` directly.

https://claude.ai/code/session_016UeeNs35PANEu6PpnmyfNM

### DIFF
--- a/frontend/src/components/RoutingGraphTypes.ts
+++ b/frontend/src/components/RoutingGraphTypes.ts
@@ -71,6 +71,7 @@ export interface RuleFlags {
     claudeCodeCompat?: boolean;
     cleanHeader?: boolean;
     context1m?: boolean;
+    claudeOrgId?: string;
 }
 
 export interface RuleFlagsApi {
@@ -88,6 +89,7 @@ export interface RuleFlagsApi {
     claude_code_compat?: boolean;
     clean_header?: boolean;
     context_1m?: boolean;
+    claude_org_id?: string;
 }
 
 export type FlagValueType = 'bool' | 'string' | 'enum' | 'int' | 'service_ref';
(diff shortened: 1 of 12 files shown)


andreasfoo (approved):
LGTM

PR #1549, opened 2026-08-11
fix(nvidia): strip top-level prompt_cache fields for NIM (Claude Code 400)

### LINKED ISSUE #1548: [BUG] NVIDIA NIM: Claude Code gets 400 "Unsupported parameter(s): prompt_cache_options"
### Version
0.260806.1

### OS & Architecture
Linux x86_64

### Category
Protocol

### Scenario
Routing real Claude Code through tingly-box to an NVIDIA NIM OpenAI-compatible endpoint (`integrate.api.nvidia.com`, model `thinkingmachines/inkling`). "Test all providers" in the UI always passes, but every real multi-turn Claude Code request fails with HTTP 400 before any generation runs.

### Steps to Reproduce
1. Add an NIM provider (`api_base = https://integrate.api.nvidia.com/v1`, `api_style = openai`) with a real NIM API key.
2. Start Claude Code pointed at the router.
3. Send a multi-turn conversation (even "hey" once CC has populated its cached system prefix is enough).

### Logs / Error Output
```
API Error: 400 Failed to create streaming request:
  POST "https://integrate.api.nvidia.com/v1/chat/completions": 400 Bad Request
```
NIM upstream body (captured via MITM on the router->NIM path):
```json
{"message":"Validation: Unsupported parameter(s): `prompt_cache_options`","type":"Bad Request","code":400}
```

### Root cause
Claude Code sends requests in **Anthropic format** (`POST .../v1/messages?beta=true`) with `cache_control: {type: ephemeral}` on its system blocks. The Anthropic->OpenAI conversion in `internal/protocol/request/anthropic_view.go` (L291-305) **derives** two top-level fields from those cache-control markers -- `openaiReq.PromptCacheOptions.Mode` (`"implicit"` / `"explicit"`) and `openaiReq.PromptCacheRetention` -- whenever any cache control is present in the request (system block, message block, tool definition, or tool use; predicates `viewHasRepresentableCacheControl` / `viewHasToolDefinitionCacheControl` / `viewHasToolUseCacheControl` / `hasAnthropicCacheControl`). It is **tingly-box, not Claude Code**, that places `prompt_cache_options` at the request top level. NIM is OpenAI-compatible but does not accept these top-level fields, and rejects the whole request.

The per-message `prompt_cache_breakpoint` content markers that tingly also emits for cache-control breakpoints **are** accepted by NIM and are not the problem -- only the two derived top-level fields are. The single-turn "Test all providers" check sets no `cache_control`, so it never trips this.

### Repro -- what Claude Code actually sends (ingress, Anthropic format)
Dumped from a live Claude Code request (`claude-cli/2.1.220`, model `tingly/cc`):
```
--- SYSTEM (3 blocks) ---
  blk[0]: cache_control=None                    len=70    text=x-anthropic-billing-header: cc_version=2.1.220.564...
  blk[1]: cache_control={'type': 'ephemeral'}   len=57    text=You are Claude Code, Anthropic's official CLI for Claude.
  blk[2]: cache_control={'type': 'ephemeral'}   len=6993  text=You are an interactive agent that helps users with software...
top-level prompt_cache_options: False   <- CC does not send this itself
```
That same request, after tingly's conversion to NIM, carries a top-level `prompt_cache_options` (derived) -> **400**. After the fix: no top-level cache fields, but the per-message `prompt_cache_breakpoint:{mode:"explicit"}` is kept -> **200**.

### Direct NIM probes (router-independent)
| Body sent to `integrate.api.nvidia.com/v1/chat/completions` | NIM response |
|---|---|
| `..., "prompt_cache_options": {"mode":"implicit"}` (top-level) | **400** `Unsupported parameter(s): prompt_cache_options` |
| `..., "prompt_cache_retention": "24h"` (top-level) | **400** `Unsupported parameter(s): prompt_cache_retention` |
| neither, messages only | **200** |
| `cache_control` / `prompt_cache_breakpoint` in message content part, neither top-level | **200** |

Stripping the two top-level fields is necessary and sufficient. A/B on the router with the same multi-turn payload: original = **400**, patched = **200**.

### Suggested fix
Add an NIM case in the `ApplyProviderTransforms` switch (`internal/protocol/ops/request_openai_extensions.go`) that strips the two derived fields before forwarding:

```go
case strings.Contains(url, "integrate.api.nvidia.com"):
    // NVIDIA NIM rejects prompt-cache fields that Claude Code sends:
    // "Unsupported parameter(s): `prompt_cache_options`". Strip them.
    return applyDefaultTransform(stripPromptCacheForNVIDIA(req), config)
```

```go
// internal/protocol/ops/request_openai_nvidia.go
func stripPromptCacheForNVIDIA(req *openai.ChatCompletionNewParams) *openai.ChatCompletionNewParams {
    req.PromptCacheOptions = openai.ChatCompletionNewParamsPromptCacheOptions{}
    req.PromptCacheRetention = ""
    return req
}
```

Both fields carry `json:"...,omitzero"` and the SDK marshals via `param.MarshalObject` (a shimmed `encoding/json` honoring `omitzero`), so zeroing them **omits the keys with no JSON round-trip**, preserving per-message extras (`x_thinking` / `reasoning_content` / `prompt_cache_breakpoint`). `applyDefaultTransform` only sets `ReasoningEffort` / `thinking` and does not re-inject the cache fields. Mirrors existing `sanitizeCodexPromptCacheJSON` in `internal/client/codex_round_tripper.go:239-240` (same strip pattern, Codex backend). Scoped to NIM: `api.openai.com` and other providers keep `prompt_cache_options`.

### Verification
- `go test ./internal/protocol/ops/` -> ok (5 new unit tests; existing deepseek/gemini untouched).
- `go test ./internal/protocol/transform/`, `./internal/protocoltest/` -> ok. `go build ./internal/...` clean.
- Live: rebuilt daemon, real `claude-cli/2.1.220` on `thinkingmachines/inkling` -> `upstream 200 via direct` (was `400`); recorded transformed body keeps the per-message `prompt_cache_breakpoint:{mode:"explicit"}` and omits top-level `prompt_cache_options` / `prompt_cache_retention`.

PawiX25:
## What
Routing Claude Code through tingly-box to NVIDIA NIM (`integrate.api.nvidia.com`, model `thinkingmachines/inkling`) fails with `400 "Unsupported parameter(s): prompt_cache_options"`. Claude Code sends an Anthropic-format request with `cache_control: {type: ephemeral}` on its system blocks; tingly's Anthropic->OpenAI conversion **derives** top-level `prompt_cache_options` / `prompt_cache_retention` from those markers, and NIM rejects both top-level fields (per-message markers are fine). The MITM-captured body sent to NIM contained a top-level `prompt_cache_options`; a direct probe of NIM confirms that field (and `prompt_cache_retention`) is the sole cause of the `400`.

## Fix
- New `stripPromptCacheForNVIDIA` zeroes `PromptCacheOptions` + `PromptCacheRetention`. Both carry `omitzero`; the SDK marshals via `param.MarshalObject` (shimmed `encoding/json` honoring `omitzero`), so the keys are omitted with no JSON round-trip -- per-message extras (`x_thinking` / `reasoning_content` / `prompt_cache_breakpoint`) are preserved. (An earlier round-trip iteration silently dropped `ExtraFields` and was rejected.)
- New `integrate.api.nvidia.com` case in `ApplyProviderTransforms`. Scoped: `api.openai.com` and other providers keep `prompt_cache_options` (covered by `TestNVIDIANonNVIDIAProvidersUnaffected`).
- `applyDefaultTransform` after the strip only sets `ReasoningEffort` / `thinking`; it does not re-inject the cache fields. Mirrors `sanitizeCodexPromptCacheJSON` in `codex_round_tripper.go:239-240`.
- Scope is NVIDIA only; other providers keep `prompt_cache_options`.

## Testing
- `go test ./internal/protocol/ops/` -> ok (5 new unit tests; existing untouched). Tests are pure unit (no HTTP); they verify `omitzero` omission + extra-field preservation + non-NVIDIA scope.
- `go test ./internal/protocol/transform/` + `./internal/protocoltest/` -> ok. `go build ./internal/...` clean.
- Direct `curl` against NIM (router-independent): `prompt_cache_options` top-level -> 400, `prompt_cache_retention` top-level -> 400, neither -> 200, per-message `prompt_cache_breakpoint` -> 200.
- 1:1 A/B on the router with the same multi-turn payload carrying `cache_control`: original binary -> 400, patched binary -> 200.
- Live: rebuilt daemon, real `claude-cli/2.1.220` on `thinkingmachines/inkling` -> `upstream 200 via direct` (was `400`); recorded transformed body keeps `prompt_cache_breakpoint:{mode:"explicit"}` on the system content part and omits top-level `prompt_cache_options`.

Fixes #1548

### DIFF
--- a/internal/protocol/ops/request_openai_extensions.go
+++ b/internal/protocol/ops/request_openai_extensions.go
@@ -26,6 +26,11 @@ func ApplyProviderTransforms(req *openai.ChatCompletionNewParams, providerURL, m
 		strings.Contains(url, "opencode.ai/zen/go") && strings.Contains(modelLower, "deepseek"):
 		return applyDeepSeekTransform(req, providerURL, model, config)
 
+	case strings.Contains(url, "integrate.api.nvidia.com"):
+		// NVIDIA NIM rejects prompt-cache fields that Claude Code sends:
+		// "Unsupported parameter(s): `prompt_cache_options`". Strip them.
+		return applyDefaultTransform(stripPromptCacheForNVIDIA(req), config)
+
 	case strings.Contains(url, "generativelanguage.googleapis.com") && strings.Contains(modelLower, "gemini"):
 		return applyGeminiTransform(req, providerURL, model, config)
 
--- a/internal/protocol/ops/request_openai_nvidia.go
+++ b/internal/protocol/ops/request_openai_nvidia.go
@@ -1,0 +1,22 @@
+package ops
+
+import (
+	"github.com/openai/openai-go/v3"
+)
+
+// stripPromptCacheForNVIDIA removes prompt-cache fields that the NVIDIA NIM
+// OpenAI-compatible endpoint rejects with:
+//
+//	400 Bad Request: {"message":"Validation: Unsupported parameter(s): `prompt_cache_options`"}
+//
+// Claude Code sends `prompt_cache_options` / `prompt_cache_retention` at the
+// request top level on every multi-turn request. Both fields are tagged
+// `omitzero` in the SDK, so zeroing them omits them from the marshaled request
+// without a JSON round-trip — which would drop per-message extra fields such
+// as `x_thinking` / `reasoning_content`.
+func stripPromptCacheForNVIDIA(req *openai.ChatCompletionNewParams) *openai.ChatCompletionNewParams {
+	req.PromptCacheOptions = openai.ChatCompletionNewParamsPromptCacheOptions{}
+	req.PromptCacheRetention = ""
+
+	return req
+}
--- a/internal/protocol/ops/request_openai_nvidia_test.go
+++ b/internal/protocol/ops/request_openai_nvidia_test.go
@@ -1,0 +1,129 @@
+package ops
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/openai/openai-go/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tingly-dev/tingly-box/internal/protocol"
+)
+
+const nvidiaNIMURL = "https://integrate.api.nvidia.com/v1"
+
+// TestNVIDIAStripPromptCacheTopLevel proves that the top-level
+// prompt_cache_options / prompt_cache_retention fields Claude Code sends are
+// removed before the request is forwarded to NVIDIA NIM.
+func TestNVIDIAStripPromptCacheTopLevel(t *testing.T) {
+	req := &openai.ChatCompletionNewParams{
+		Model:                openai.ChatModel("thinkingmachines/inkling"),
+		PromptCacheOptions:   openai.ChatCompletionNewParamsPromptCacheOptions{Mode: "implicit"},
+		PromptCacheRetention: openai.ChatCompletionNewParamsPromptCacheRetention("1400h"),
+		Messages: []openai.ChatCompletionMessageParamUnion{{
+			OfUser: &openai.ChatCompletionUserMessageParam{
+				Name:    openai.String("u"),
+				Content: openai.ChatCompletionUserMessageParamContentUnion{OfString: openai.String("hi")},
+			},
+		}},
+	}
+
+	ApplyProviderTransforms(req, nvidiaNIMURL, string(req.Model), &protocol.OpenAIConfig{})
+
+	raw := marshalParams(t, req)
+	assert.NotContains(t, raw, "prompt_cache_options",
+		"prompt_cache_options must be stripped for NVIDIA NIM")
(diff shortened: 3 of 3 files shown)


FFengIll (approved):
Approved, and we may think twice for prompt_cache option / config and do some change in future .

0x0079:
I believe prompt_cache is not feasible for most providers, since it is a new prompt_cache after gpt 56 sol, but not a stable one like anyother schema.

andreasfoo:
OK for NVIDIA. We could generalize this to other providers as well, but we’re not sure yet which ones should follow the same approach. So perhaps we can merge the NVIDIA-specific version first and generalize it later once we have a better understanding.

PR #1553, opened 2026-08-12
perf(db): cache hot-path reads, merge stats+usage writes

0x0079:
## Summary

pprof on the real routing pipeline (`internal/routing/pipeline_bench_test.go`) showed `ServiceSelector.Select` spending 47.5% of its CPU in `ProviderStore.GetByUUID` — a SQLite round-trip on every routed request, serialized behind a plain `sync.Mutex` that blocked concurrent readers too. `APITokenStore.ValidateToken` (auth middleware) and `StatsStore`/`UsageStore`'s per-request writes had the same shape. Every fix below was benchmarked before/after; see `.design/hot-path-db-access.md` for the full writeup, including a rejected approach (`PRAGMA synchronous=NORMAL` — the driver already applies it, so it measured as a no-op) kept there so it isn't retried.

- **`ProviderStore`/`APITokenStore`**: write-through in-memory cache instead of a SQLite round-trip per read. `BenchmarkSelect_Plain` 59.7µs→5.8µs/op (10.2×), `ValidateToken` 30.0µs→119ns/op (~251×). `UpdateLastUsed` debounced to once/minute per token, taking the full auth middleware path to ~75× total.
- **`StatsStore`/`UsageStore`**: both are per-request writes, not reads, so no cache applies — `RecordRequestOutcome` merges their two independent transactions into one (~4–8%, plus atomicity: the stats and usage rows can no longer partially persist).
- Fixed a latent bug found along the way: `NewUsageStore` never created its `db` subdirectory (every other `NewXStore` does) — never hit in production since `StoreManager` never calls it directly.
- Code review + a 4-angle simplify pass on top: fixed two real bugs (cache mutated before a SQLite write succeeded; a nil-usage fallback that surfaced a spurious error), removed a speculative-generality fallback branch three independent reviewers flagged, and factored duplicated cache-write/store-construction logic into shared helpers.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (whole repo, green throughout)
- [x] Every perf claim backed by a `go test -bench` run before/after (tables in `.design/hot-path-db-access.md`)
- [x] Regression tests: cache-not-corrupted-on-failed-write, atomic transaction rollback, debounce behavior

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_015x1eojavC1neC8U2GnLzLo

### DIFF
--- a/.design/hot-path-db-access.md
+++ b/.design/hot-path-db-access.md
@@ -1,0 +1,110 @@
+# Hot-Path DB Access: Cache Reads, Don't Lock+Query Them
+
+Status: `ProviderStore` fixed on branch `claude/routing-pipeline-benchmark-e6aq73`. Rest of the pattern below is diagnosed but **not yet fixed** — tracked here as the holistic follow-up.
+
+## Why
+
+`internal/routing/pipeline_bench_test.go` benchmarks the real production selection
+pipeline (`ServiceSelector` + `protocolserver.LoadBalancer`, wired exactly as
+`internal/server/server.go` does). Profiling it (`go test ./internal/routing/...
+-bench . -cpuprofile=cpu.prof`) showed `ServiceSelector.Select` spending **83% of
+its own time** in one line:
+
+```go
+provider, err := s.config.GetProviderByUUID(result.Service.Provider)
+```
+
+That call chain — `config.Config.GetProviderByUUID` → `db.ProviderStore.GetByUUID`
+→ `gorm.DB.First` → SQLite via cgo — accounted for **47.5% of the entire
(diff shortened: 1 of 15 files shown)


0x0079:
Splitting this into independent PRs given the scope:

- #1556 — docs(db): add db.md with a Performance section
- #1557 — perf(db): cache ProviderStore reads
- #1558 — perf(db): cache+debounce APITokenStore
- #1559 — perf(db): merge StatsStore+UsageStore writes into one transaction

Closing this one in favor of those.

---

PR #1554, opened 2026-08-12
fix(protocol): default-deny OpenAI prompt-cache fields per vendor

### LINKED ISSUE #1548: [BUG] NVIDIA NIM: Claude Code gets 400 "Unsupported parameter(s): prompt_cache_options"

0x0079:
## Summary

OpenAI's gpt-5.6+ prompt-cache fields (`prompt_cache_options` / `prompt_cache_retention` / `prompt_cache_breakpoint`) were emitted to every OpenAI-compatible vendor, though most don't implement them and strict-schema gateways 400 on the unknown field — exactly the NVIDIA NIM failure in #1548. This flips the default to strip, with a per-vendor allowlist for the vendors confirmed to accept them.

## Key Changes

- **Safe-by-default vendor compatibility**: `ApplyProviderTransforms` strips all prompt-cache fields unless the destination is on the `supportsExplicitPromptCache` allowlist (currently `api.openai.com`).
- **NIM fix generalized**: folds #1549's `integrate.api.nvidia.com` special case into the default path, also absorbing its `prompt_cache_retention` strip; the #1548 regression tests are kept green unchanged.
- **Vendor-dispatch harness**: new `harness matrix --mode=vendor` suite routes real vendor `APIBase` strings (openai/deepseek/nvidia/generic) through the live gateway to pin allowlist behavior end-to-end.
- **cache_controls expectations**: the existing suite's generic providers are correctly off the allowlist, so its assertions now expect stripping on Chat-shape hops (`cacheSurvivesPath`).

## Notes

- Onboarding a vendor that supports these fields = one line in `supportsExplicitPromptCache`; NIM-style strict vendors need no code at all.

Fixes #1548 · Supersedes the vendor-specific approach of #1549


0x0079:
Rebased onto main, which brought in #1549 (the NVIDIA NIM fix for #1548). NIM's `400 "Unsupported parameter(s): prompt_cache_options"` is a concrete instance of exactly the failure mode this PR's default-deny generalizes, so the rebase folds the two together rather than keeping both mechanisms:

- `stripOpenAIPromptCacheFields` now also zeroes `PromptCacheRetention` — the one field #1549 stripped that the general path missed (Claude Code sends it top-level alongside `prompt_cache_options`).
- The `integrate.api.nvidia.com` dispatch case and `stripPromptCacheForNVIDIA` are removed: NIM is simply not on the `supportsExplicitPromptCache` allowlist, so the default strip covers it with no vendor-specific branch.
- The #1548 regression tests are kept unchanged (plus a header note pointing at the general mechanism) — they assert through `ApplyProviderTransforms` and stay green. Their non-NVIDIA control happens to be `api.openai.com`, the allowlisted vendor, so they also pin the allowlist's positive case.
- The protocoltest vendor suite gains an `nvidia_nim` fixture (now 8 cases, all passing), pinning the real-world case end-to-end.

Net effect: any future vendor with NIM-style strict schema validation is safe by default — no per-vendor strip needed unless a vendor is being *added to* the allowlist.

---

FFengIll (approved):

PR #1555, opened 2026-08-12
fix(quota): tolerate transient refresh failures

E-lai:
## Summary

- Retry quota fetches only for transient transport failures (EOF, unexpected EOF, connection errors, and timeouts) with bounded exponential backoff and jitter
- Keep the last successful quota snapshot when a refresh fails, while recording the refresh error and scheduling a near-term retry
- Coalesce concurrent refreshes for the same provider without coupling callers' contexts
- Avoid overwriting cached quota when persistence reads are uncertain or the request is canceled

## Test plan

- [x] `go test ./...` (`ai` module)
- [x] `go test -race ./quota/...`
- [x] `go vet ./quota/...`
- [x] `go build ./...`
- [x] `go test ./internal/server/module/providerquota/...`

### DIFF
--- a/ai/quota/manager.go
+++ b/ai/quota/manager.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"math/rand/v2"
 	"net"
 	"net/url"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


0x0079:
Thanks for the PR.

As an independent module for  ai provider quota, it may be too complex.

Is there a very strong motivation or blocking problem for this PR?

FFengIll:
Temporarily closed due to lack of response.

Issue #1611, opened 2026-08-20
[TEP-1611] Multiple Team Management with sharing key


FFengIll:
### Category

Agent Management

### Motivation & Problem Statement

Like claude code profile, we require multiple team management to isolate models and keys.

0x0079:
#1630 #1631

PR #1628, opened 2026-08-24
feat(server,frontend): add multi-team workspaces for sharing-key routing isolation

FFengIll:
## Summary
Sharing keys previously all shared one global routing scope, so there was no way to isolate one integration's traffic and configuration from another's. This adds first-class teams: each team gets its own routing scope, sharing keys are bound to (and movable between) teams, and requests  are auto-scoped server-side from the authenticated credential.

It implements #1611.

## Key Changes

- **Team-scoped routing**: requests authenticate to a specific team's rule set via `teamScopeMiddleware`, derived from the credential, never from client input.
- **Team lifecycle management**: new `internal/server/module/team` APIs to create, rename, enable/disable, and delete teams, with a stable `default` team preserving legacy behavior.
- **Sharing keys bound to teams**: keys are created under a team and can be moved between teams; disabling a team immediately invalidates its keys.
- **Multi-team workspace UI**: new team page, sidebar navigation, and sharing-keys dialog with team-scoped create/move/delete flows.
- **Sharing-key expiry enforced**: `ValidateToken` now actually rejects tokens past `ExpiresAt`, closing a pre-existing gap surfaced while scoping keys to teams.

## Minor
- Deduped repeated error-status inference, action boilerplate, and ID-generation logic uncovered while building the feature (team handler, `UseTeamPage`, `SharingKeysDialog`, team-slug generation).
- Fixed a UI race where switching teams quickly could briefly show another team's sharing keys, and split team/profile creation's shared loading flag.

## Notes
- `internal/server/module/sharing`'s `GET /tokens` still returns raw secrets with no forced team filter — worth a follow-up ticket, out of scope here.


FFengIll:
close since #1630 #1631

PR #1633, opened 2026-08-25
fix(visionproxy): describe images in the tool channel, not just user messages

cbrown350:
## Summary

Since #1609 widened tool-result content to the full part union, tool-returned images reach the provider untouched, so a text-only model is handed an image it cannot read instead of a description.

## Key Changes

- **Tool-returned screenshots**: `collectOpenAI` walks tool messages alongside user messages, so their images are described like any other.
- **Text-only providers**: z.ai/glm-5.3 went from `400 code 1210 "messages.content.type is invalid, allowed values: ['text']"` to answering from the injected description.
- **Test blind spot**: the shared count/collect helpers were user-only too, and now read parts role-agnostically, deliberately separate from `collectOpenAI`.

### DIFF
--- a/internal/vision/visionproxy/vision_proxy.go
+++ b/internal/vision/visionproxy/vision_proxy.go
@@ -300,18 +300,25 @@ func collectV1(req *anthropic.MessageNewParams) []imageRef {
 	return refs
 }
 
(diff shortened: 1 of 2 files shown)


FFengIll:
Thanks for the fix.

PR #1634, opened 2026-08-25
fix(claude): TitleCase all OAuth tool names and restore them on streams

cbrown350:
## Summary

Anthropic's OAuth path rejects requests carrying many snake_case tool names and bills them to Extra Usage rather than the subscription. Renaming only the fourteen tools in the known-name map left any client with a toolset of its own failing with `You're out of extra usage`.

## Key Changes

- **Subscription billing on OAuth**: every custom tool name folds to TitleCase, not only the fourteen known ones, so large toolsets stay on the subscription quota.
- **Streaming tool dispatch**: `MessagesNewStreaming` discarded the reverse map, so renamed tools returned TitleCased and callers could not match them; the rename is now undone on the wire.
- **Rename collisions**: a fold that would collide with another tool in the same request is skipped, leaving both names untouched.
- **`tool_choice` / history consistency** (added after review): a pinned tool and prior-turn `tool_use` blocks now follow the same rename plan as `tools[]`, instead of keeping their original name and desyncing from it.
- **MCP tool names** (added after review): `mcp__<server>__<tool>` and Tingly-Box's own `tingly_box_mcp__<source>__<tool>` are now left untouched end-to-end, matching what Claude Code actually sends.

## Notes

- Threshold measured against `api.anthropic.com`: sixteen snake_case names are rejected, fifteen accepted, twenty-nine TitleCased accepted.
- It is a casing heuristic, not an allowlist — an unknown `BrowserExec` passes.

### DIFF
--- a/internal/client/claude_tool_names.go
+++ b/internal/client/claude_tool_names.go
@@ -12,10 +12,20 @@ import (
 	anthropicOption "github.com/anthropics/anthropic-sdk-go/option"
 )
 
-// mcpToolPrefix marks MCP-provided tools. Claude Code keeps the prefix intact
-// and only capitalizes the first character of the remainder, so we match that
-// rather than folding the whole name.
-const mcpToolPrefix = "mcp_"
+// mcpNamespaceSeparator separates the segments of an MCP tool name.
+//
+// Claude Code passes MCP tools through verbatim as mcp__<server>__<tool> —
+// lowercase, unfolded — so a name already using this separator is left alone.
+// Folding it would move the request away from real Claude Code traffic rather
+// than towards it, and would flatten the namespace the client uses to route the
+// call. Tingly-Box's own server tools follow the same convention
+// (tingly_box_mcp__<source>__<tool>, see internal/tool.NormalizeToolName), so
+// one rule covers both.
+//
+// A single-underscore "mcp_" prefix is deliberately *not* treated as a
+// namespace: it is not a shape any first-party client emits, so it folds like
+// any other snake_case name.
+const mcpNamespaceSeparator = "__"
 
 // claudeCodeToolName returns the name Claude Code would send for a tool.
 //
@@ -27,10 +37,14 @@ const mcpToolPrefix = "mcp_"
 // fail while the same request with TitleCased names succeeds; the check is a
 // casing heuristic, not an allowlist of known Claude Code tools.
 //
-// oauthToolRenameMap still wins for the well-known Claude Code tools so those
+// MCP-namespaced names are exempt — see mcpNamespaceSeparator.
+// oauthToolRenameMap then wins for the well-known Claude Code tools so those
 // keep their exact official spelling (e.g. "ls" -> "LS", not "Ls"). Everything
 // else falls back to a mechanical TitleCase.
 func claudeCodeToolName(name string) string {
+	if strings.Contains(name, mcpNamespaceSeparator) {
+		return name
+	}
 	if mapped, ok := oauthToolRenameMap[name]; ok {
 		return mapped
 	}
@@ -42,12 +56,6 @@ func claudeCodeToolName(name string) string {
 func titleCaseToolName(name string) string {
 	if name == "" {
 		return name
-	}
-	if rest, ok := strings.CutPrefix(name, mcpToolPrefix); ok {
-		if rest == "" {
-			return name
-		}
-		return mcpToolPrefix + upperFirst(rest)
 	}
 	var b strings.Builder
 	b.Grow(len(name))
--- a/internal/client/claude_tool_names.go
+++ b/internal/client/claude_tool_names.go
@@ -29,13 +29,14 @@ const mcpNamespaceSeparator = "__"
 
 // claudeCodeToolName returns the name Claude Code would send for a tool.
 //
-// Anthropic fingerprints Claude Code OAuth traffic partly on tool naming: a
-// request carrying many lowercase/snake_case tool names is classified as
-// third-party and billed to Extra Usage rather than the subscription, which
-// surfaces as a 400 "You're out of extra usage" once that bucket is spent.
-// Measured against api.anthropic.com, requests with >=16 snake_case tool names
-// fail while the same request with TitleCased names succeeds; the check is a
-// casing heuristic, not an allowlist of known Claude Code tools.
+// Anthropic classifies some OAuth traffic as third-party and bills it to Extra
+// Usage rather than the subscription, which surfaces as a 400 "You're out of
+// extra usage" once that bucket is spent; tool naming appears to be one of the
+// signals. An earlier report measured a hard cutoff (~16 snake_case names);
+// later probes — five arms up to 48 lowercase names at 100% density — did not
+// reproduce it, so the exact trigger is unconfirmed (see PR #1634). The fold
+// is kept regardless because it only moves names towards the shape Claude Code
+// itself sends, and at worst is a no-op.
 //
 // MCP-namespaced names are exempt — see mcpNamespaceSeparator.
 // oauthToolRenameMap then wins for the well-known Claude Code tools so those
--- a/internal/client/claude_tool_names.go
+++ b/internal/client/claude_tool_names.go
@@ -1,0 +1,219 @@
+package client
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+	"unicode"
+
+	anthropicOption "github.com/anthropics/anthropic-sdk-go/option"
+)
+
+// mcpToolPrefix marks MCP-provided tools. Claude Code keeps the prefix intact
+// and only capitalizes the first character of the remainder, so we match that
+// rather than folding the whole name.
+const mcpToolPrefix = "mcp_"
+
+// claudeCodeToolName returns the name Claude Code would send for a tool.
+//
+// Anthropic fingerprints Claude Code OAuth traffic partly on tool naming: a
+// request carrying many lowercase/snake_case tool names is classified as
+// third-party and billed to Extra Usage rather than the subscription, which
+// surfaces as a 400 "You're out of extra usage" once that bucket is spent.
+// Measured against api.anthropic.com, requests with >=16 snake_case tool names
+// fail while the same request with TitleCased names succeeds; the check is a
+// casing heuristic, not an allowlist of known Claude Code tools.
+//
+// oauthToolRenameMap still wins for the well-known Claude Code tools so those
+// keep their exact official spelling (e.g. "ls" -> "LS", not "Ls"). Everything
+// else falls back to a mechanical TitleCase.
+func claudeCodeToolName(name string) string {
+	if mapped, ok := oauthToolRenameMap[name]; ok {
+		return mapped
+	}
+	return titleCaseToolName(name)
+}
+
+// titleCaseToolName folds a snake_case tool name into TitleCase.
+// Names that are already TitleCase come back unchanged.
+func titleCaseToolName(name string) string {
+	if name == "" {
+		return name
+	}
+	if rest, ok := strings.CutPrefix(name, mcpToolPrefix); ok {
+		if rest == "" {
+			return name
+		}
+		return mcpToolPrefix + upperFirst(rest)
+	}
+	var b strings.Builder
+	b.Grow(len(name))
+	for _, part := range strings.Split(name, "_") {
+		if part == "" {
+			continue
+		}
+		b.WriteString(upperFirst(part))
+	}
+	if b.Len() == 0 {
+		return name
+	}
+	return b.String()
+}
+
+func upperFirst(s string) string {
+	if s == "" {
+		return s
+	}
+	r := []rune(s)
+	r[0] = unicode.ToUpper(r[0])
+	return string(r)
+}
+
+// planToolRenames maps original tool name -> outbound name for a request.
+//
+// A rename is skipped when the target collides with another tool in the same
+// request (either one already named that, or another rename targeting it).
+// Two distinct tools sharing one outbound name would make the reverse mapping
+// ambiguous and could dispatch a tool result to the wrong tool, so leaving the
+// colliding pair untouched is the safe failure mode.
+func planToolRenames(names []string) map[string]string {
+	taken := make(map[string]struct{}, len(names))
+	for _, n := range names {
+		taken[n] = struct{}{}
+	}
+	plan := make(map[string]string, len(names))
+	used := make(map[string]struct{}, len(names))
+	for _, n := range names {
+		newName := claudeCodeToolName(n)
+		if newName == n {
+			continue
+		}
+		if _, clash := taken[newName]; clash {
+			continue
+		}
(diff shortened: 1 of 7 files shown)


internal/client/claude_tool_names.go
```diff
@@ -0,0 +1,219 @@
+package client
```
0x0079, line 1: @andreasfoo I think you have checked this before (including name mapping).
Please help check this PR.

@FFengIll Is it a protocol spec or a client spec?

cbrown350: Not aiming to answer for @FFengIll, but since we did the measurements — as far as we can tell it's neither. The Messages API documents no casing constraint on tool names (Anthropic's own tool-use examples are snake_case), and Claude Code doesn't publish a naming rule either; its built-ins just happen to be TitleCase.

What we can observe is a server-side heuristic on the OAuth path, keyed on casing and count rather than identity: 15 snake_case names accepted, 16 rejected, 29 TitleCased accepted, and an unknown `BrowserExec` passes — so it's not an allowlist. The failure mode is billing rather than a hard error: rejected requests come back `You're out of extra usage`, having been pushed off the subscription onto Extra Usage.

Caveat: measured on the OAuth path only; I haven't checked whether an API key behaves the same. Since there's no documented contract here, this PR keeps the fold scoped to that path rather than normalizing names generally — happy to reshape it if you'd rather it live elsewhere.

andreasfoo:
Thanks for digging into this — the diagnosis is solid, and the streaming find is a genuine bug (dropping reverseMap in MessagesNewStreaming leaves the client unable to match tool names). I also like that collisions leave both names alone rather than risking a mis-dispatch. 

Two things I think need a look before merge:

- tool_choice isn't remapped. remapToolNames rewrites req.Tools but not req.ToolChoice, so a pinned tool ends up pointing at a name that no longer exists:

  tools[0]="ReadFile"   tool_choice.name="read_file"

  Clients reach this through openai_to_anthropic.go:275 and google_to_any.go:407, as well as natively. It was mostly harmless before, since only the fourteen known lowercase names could trigger it — but with every tool folding, any request that pins a tool now hits it. Applying the same plan to ToolChoice.OfTool.Name should cover it.

-  MCP tools don't get folded. mcpToolPrefix is mcp_, but real Claude Code MCP names use a double underscore, so CutPrefix leaves a leading _ that upperFirst can't touch:

  mcp__github__get_pr  ->  mcp__github__get_pr   (unchanged)

  That means a client with a dozen-plus MCP tools — the case most likely to cross the 16-name threshold — comes through with nothing renamed. Might be worth re-checking the comment's claim about Claude Code capitalizing the remainder, too; my understanding is those names stay lowercase.

cbrown350:
Both confirmed and fixed — thanks, these were real.

Before touching anything I reproduced the `tool_choice` case against the live API, and Anthropic rejects it outright:

```
POST https://api.anthropic.com/v1/messages → 400
{"type":"invalid_request_error",
 "message":"Tool 'foo__bar' not found in provided tools"}
req_011CeTiTYki2vADMWgKsbUDD
```

`tools[]` carried the folded name, `tool_choice` still carried the original. Exactly the desync you described — it's a hard request failure, not a subtle one.

**tool_choice.** Fixed, and the same argument extends to a third site: `tool_use` blocks in prior assistant turns. Those carry the name the model was given last time, so every follow-up turn in a tool conversation was re-sending the original snake_case names — putting back on the wire exactly what this rename exists to remove. Same escalation you identified for `tool_choice`: mostly harmless when only fourteen known names could fold, unavoidable now that everything does, and it compounds with conversation length.

Rather than patch each site, I changed the signature. `remapToolNames(tools)` structurally couldn't reach the rest of the request; it's now `remapRequestToolNames(req)`, computing the plan once and applying it to `tools`, `tool_choice`, and history together. That single entry point matters for the reason you'd expect — `planToolRenames` deliberately skips a fold whose target collides, so folding `tool_choice` independently could pin a name that isn't in `tools[]`, or one that belongs to a different tool.

**MCP.** Your read of the mechanism is exactly right: `CutPrefix` leaves a leading `_` that `upperFirst` can't touch, so `mcp__github__get_pr` came through untouched. I went the other way on the fix, and your last sentence is why — those names *do* stay lowercase. Claude Code passes MCP tools through verbatim, so folding them moves the request away from real Claude Code traffic rather than towards it, and flattens the namespace the client routes on. They're now exempt deliberately instead of by accident.

Checking that turned up the inverse bug. Tingly-Box's own server tools are `tingly_box_mcp__<source>__<tool>` (see `internal/tool.NormalizeToolName`) and don't start with `mcp_`, so they skipped that branch entirely and took the full fold:

```
mcp__github__get_pull_request             ->  mcp__github__get_pull_request     (untouched)
tingly_box_mcp__webtools__mcp_web_search  ->  TinglyBoxMcpWebtoolsMcpWebSearch  (mangled)
```

Same `__` convention on both, so one rule now covers them.

**The comment.** You were right to doubt it. "Keeps the prefix intact and only capitalizes the first character of the remainder" describes a third-party OAuth shim's transform, not Claude Code. Removed rather than corrected.

**One loose end, offered as an observation rather than a conclusion.** I tried to measure whether lowercase `mcp__…` names count toward the ~16-name threshold, and couldn't get the threshold to fire at all. Testing with Extra Usage at $0 — where a third-party classification hard-fails instead of silently billing, so a 200 is a real negative — none of these tripped it:

| lowercase names on the wire | density | result |
|---|---|---|
| 20 `mcp__…` | 100% | 200 |
| 40 classic `foo_bar` | 50% | 200 |
| 40 single-underscore, no doubles | 80% | 200 |
| 48 (24 single-underscore), none folded | 100% | 200 |

That points away from a count-based rule at 16 and away from a pure ratio rule, but I'd want someone to confirm it independently before anyone leans on it — my arms had to smuggle names past `planToolRenames` to keep them unfolded, and I can't rule out that the classifier keys on something my constructions don't reproduce, or that this varies by account state. Flagging it because it seems worth knowing, not because I think it settles anything. It doesn't affect these fixes either way: the exemption keeps MCP names in the shape Claude Code actually sends.

**Tests.** Unit coverage per site, plus httptest wire tests that assert the serialized body Anthropic would receive and the stream read back. All eight were confirmed failing against the un-fixed code before the fix went in.

0x0079:
@codex read anthropic golang sdk from our tingly-dev fork version, find out why / how

0x0079:
@codex read anthropic golang sdk from our tingly-dev fork version, and check around this issue

0x0079:
## Independent review: analysis, SDK-level verification, and verdict

I went through the diff, then went and read the pinned `anthropic-sdk-go` fork (`tingly-dev-v1.66.0`) directly to check whether the architecture this PR relies on (an HTTP middleware rewriting SSE bodies) actually holds up, rather than taking the design rationale on faith. Summary: **the fix is correct and the SDK confirms the approach is necessary, not just convenient.**

### 1. What actually got fixed, across both commits

| # | Bug | Where it lived | Fix |
|---|-----|-----------------|-----|
| 1 | Only the 14 known Claude Code tool names got TitleCased; any client with its own tool vocabulary stayed over Anthropic's ~16-name snake_case threshold and got billed to Extra Usage | `oauthToolRenameMap` lookup only | Unknown names now fold mechanically to TitleCase too (`claudeCodeToolName` → `titleCaseToolName` fallback), map still wins for official spellings (`ls`→`LS`, not `Ls`) |
| 2 | Streaming responses never got renamed back | `MessagesNewStreaming`/`BetaMessagesNewStreaming` discarded `reverseMap` | `restoreToolNamesMiddleware` rewrites `content_block_start` events on the wire |
| 3 (review round 1) | `tool_choice` pinned a name that no longer existed once `tools[]` was folded → hard 400 from Anthropic (`Tool 'foo__bar' not found`) | `remapToolNames(tools)` had no access to the rest of the request | Signature changed to `remapRequestToolNames(req)`; `ToolChoice.OfTool` now follows the same plan |
| 4 (review round 1) | MCP tool names handled backwards: `mcp__github__get_pr` fell through untouched by accident (leading `_` left by `CutPrefix`), while Tingly-Box's own `tingly_box_mcp__...` tools got fully mangled | Ad-hoc `mcp_` prefix check | Single rule: any name containing `__` is treated as a namespace and left alone |
| 5 (review round 1) | Prior-turn `tool_use` blocks in conversation history still carried the client's original snake_case name, re-introducing exactly what the rename removes, worse the longer the conversation | not handled at all | History `tool_use.Name` now folds through the same plan as `tools[]` |

### 2. Concrete before/after example

Given this request (mixed toolset, a pinned tool, and one prior turn):

```json
{
  "tools": [
    {"name": "read_file"},
    {"name": "mcp__github__get_pull_request"}
  ],
  "tool_choice": {"type": "tool", "name": "read_file"},
  "messages": [
    {"role": "assistant", "content": [
      {"type": "tool_use", "name": "read_file", "id": "toolu_0"}
    ]}
  ]
}
```

**Before this PR** (post-fold, pre-fix): `tools[0].name` becomes `ReadFile`, but `tool_choice.name` and the history `tool_use.name` are still `read_file` — the request is internally inconsistent and Anthropic rejects it (400, confirmed against the live API per the PR author's repro in the thread).

**After this PR**, everything in the request agrees:

```json
{
  "tools": [
    {"name": "ReadFile"},
    {"name": "mcp__github__get_pull_request"}
  ],
  "tool_choice": {"type": "tool", "name": "ReadFile"},
  "messages": [
    {"role": "assistant", "content": [
      {"type": "tool_use", "name": "ReadFile", "id": "toolu_0"}
    ]}
  ]
}
```

...and on the way back, the SSE stream's `content_block_start` for a `ReadFile` tool_use is rewritten back to `read_file` before it reaches the client, via `reverseMap = {"ReadFile": "read_file"}`. `mcp__github__get_pull_request` is left untouched end-to-end since it matches the `__` namespace rule. This full round trip is pinned by `TestWire_BetaStreaming_RenamesEverySiteAndRestoresTheStream`.

### 3. What I verified in `libs/anthropic-sdk-go` (the "does this architecture actually work" question)

The design rationale (in the PR discussion, though see caveat below) claims there's no per-request, decoded-event transform hook in the SDK, so the rename has to be undone at the HTTP layer via `option.WithMiddleware` before the SSE decoder is built. I checked this directly against the fork rather than trusting the claim:

| Question | Finding | Citation |
|---|---|---|
| Does streaming actually go through the same middleware chain as non-streaming? | Yes — both `MessagesNewStreaming` and `BetaMessagesNewStreaming` call the identical `requestconfig.ExecuteNewRequest` path; when the response target is `**http.Response`, `Execute()` returns the middleware-modified response directly, skipping JSON decode | `message.go:86`, `betamessage.go:95`, `internal/requestconfig/requestconfig.go` |
| Could the middleware see gzip-compressed bytes instead of plaintext? | No — the SDK never sets `Accept-Encoding`/`DisableCompression` anywhere, so Go's `http.Transport` does transparent gzip decoding before any middleware runs | grepped `default_http_client.go`, `requestconfig.go`, `option/requestoption.go`, `client.go` — zero hits |
| Does `ssestream` provide a better hook (e.g. a registered per-event decoder) that this PR should have used instead? | No — `ssestream.NewDecoder` just wraps `res.Body` in a `bufio.Scanner`; the only decoder registry (`ssestream.RegisterDecoder`) is never called anywhere in the repo, and the one custom decoder that *does* exist (`apijson.RegisterCustomDecoder`) is request-side only (normalizes string content into `TextBlockParam`) and has nothing to do with response tool names | `packages/ssestream/ssestream.go:24-38,77-79`, `internal/apijson/registry.go:56` |
| Do multiple middlewares / retries interact badly? | Middlewares chain in registration order and are rebuilt once per `Execute()` call, but the composed chain re-runs on every retry attempt. Not an issue here since `ClaudeClient` sets `WithMaxRetries(0)` — worth re-checking if that ever changes | `option/requestoption.go:196-205`, `internal/requestconfig/requestconfig.go:419-453` |

Conclusion: the middleware-based fix isn't a workaround — given this SDK's actual internals, it's the *only* place a request-scoped, collision-aware rename map can be undone before the client sees the stream.

### 4. Minor, non-blocking observations

- Two tools sharing the exact same original name in one request would leave `planToolRenames`'s collision map slightly inconsistent (only one gets folded). Very low real-world risk since Anthropic already requires unique tool names per request.
- The `__`-anywhere heuristic for "this is an MCP namespace" is intentionally blunt — a hypothetical custom tool with a stray double underscore would be exempted from folding too. Explicitly called out and tested (`TestClaudeCodeToolName_FoldsSingleUnderscoreMcpNames` vs the namespace tests), so this is a deliberate tradeoff, not an oversight.
- The two `chatgpt-codex-connector[bot]` comments reference a `.design/anthropic-sdk-stream-tool-name-restoration.md` doc and commits `99a6481`/`5db4dde` — those never actually reached this PR (that sandbox had no git remote configured, per its own "Testing" notes), so despite being referenced in this thread, that documentation doesn't currently exist on this branch. Worth a follow-up if that write-up is still wanted.
- Only one check run is visible on the head commit (`Compile gui/wails3 (Linux)`, success) and `get_status` returns no statuses — I couldn't confirm from CI that `go test ./internal/client` (the new unit + wire tests) has actually run, only that the author reports running it locally.

---

0x0079 (approved):
Approved, and thanks to the report and completed test.

0x0079:
@cbrown350 If possible, please simplify the tests before merge.

cbrown350:
Done — tests cut from ~900 to ~560 lines across the four files, no behavior change. What went and why:

- **Beta mirror suites.** Every `TestRemap{Request,Beta}RequestToolNames` subtest existed twice, once per request type. The v1 types keep the full unit coverage; the Beta twin now has one smoke test here plus the end-to-end wire test, since the Beta path is fully pinned there.
- **Remap suite was split across three files** (`claude_round_tripper_test.go`, `claude_client_test.go`, `claude_tool_names_test.go`); consolidated into `claude_tool_names_test.go` where the rest of the naming tests live. The v1/Beta wire tests also kept a full and a compact copy of the same assertions; the compact twin now just proves the v1 path with the redundant assertions removed.
- **SSE rewriter:** five passthrough subtests collapsed into one table; the multi-event + tiny-reads case absorbed the rewrite subtest, so framing and partial-read behavior are still exercised.
- **Docs.** Your comment noting the missing `.design/anthropic-sdk-stream-tool-name-restoration.md` — it never existed on this branch, it was written in a sandbox with no remote. It exists now, and covers the SDK reasoning from your review (why streaming restore has to live at the HTTP layer given this SDK's internals). While in there I also softened the doc comment on `claudeCodeToolName`: it stated the ~16-name threshold as measured fact, which my own probes in this thread failed to reproduce — it now says the trigger is unconfirmed and points at this PR.

On your two minor observations: agreed on both, no change made. The duplicate-name collision is pre-existing `planToolRenames` behavior that Anthropic's own uniqueness requirement makes unreachable in practice, and the blunt `__` heuristic is the deliberate tradeoff — the namespace flattening it prevents is worse than a hypothetical stray double underscore. Also noted and dismissed your retry question: `WithMaxRetries(0)` on this client, confirmed in your table too.

All tests pass, `go vet` clean.

PR #1635, opened 2026-08-25
build(docker): make the source image buildable and fork-pushable

cbrown350:
## Summary

`docker.build.Dockerfile` could not produce a working image and no workflow referenced it, so there was no path from repository source to a running container.

## Key Changes

- **Source builds compile**: the builder moves to `golang:1.26-alpine`, matching the `go 1.26` that go.mod already requires.
- **Dashboard reachable**: the web UI is built and embedded, without which the binary compiles and then returns HTTP 500 for every UI route.
- **Fork registries**: the image name follows `github.repository_owner` rather than a hardcoded namespace, and a dispatchable workflow builds any ref with submodules initialised.

### DIFF
--- a/.github/actions/docker-build-push/action.yml
+++ b/.github/actions/docker-build-push/action.yml
@@ -50,7 +50,7 @@ runs:
       id: meta
       uses: docker/metadata-action@v5
       with:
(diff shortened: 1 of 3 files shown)


0x0079 (approved):
LGTM, and it is helpful if a run log as attachment.

PR #1637, opened 2026-08-25
fix(visionproxy): stop silently dropping images a reasoning model describes

cbrown350:
## Summary

The describe call was capped at a hardcoded 256 tokens. A reasoning model spends that on its chain of thought before writing any content, so the proxy collects an empty string, logs `empty description; stripping image`, and the request continues without the image — the caller is never told, and answers as though none was sent.

## Key Changes

- **Budget no longer binds**: the cap defaults to 4096 and reads `TINGLY_VISION_MAX_TOKENS`; it exists to satisfy Anthropic's required `max_tokens`, not to control cost, which is decided by the model the proxy points at.
- **Exhausted budget is reported**: hitting the cap with nothing written now returns an error naming the model, the budget, and the variable to raise, instead of being folded into the generic empty-description path.
- **Truncated descriptions are reported**: a description cut off mid-sentence is still used, but no longer passes unremarked — a 1-token cap previously yielded `description=A` with nothing logged.

## Notes

- Measured against qwen3.5:397b describing a photograph: 256 produced zero content tokens on every attempt, 1024 a usable description on 4 of 5, 2048 on all 5.
- Only the empty case is fixed by raising the budget; a genuinely empty answer stays a non-error strip, as before.

### DIFF
--- a/internal/vision/visionproxy/vision_proxy_client.go
+++ b/internal/vision/visionproxy/vision_proxy_client.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 
 	"github.com/anthropics/anthropic-sdk-go"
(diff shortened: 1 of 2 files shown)


internal/vision/visionproxy/vision_proxy_client.go
```diff
@@ -48,7 +50,62 @@ type poolVisionClient struct {
...
+// tokens on every attempt; 1024 succeeded on 4 of 5, 2048 on all 5. The
+// prompt already asks for a concise answer, which is what actually keeps the
+// response short.
+const defaultVisionMaxTokens = 4096
+
+// visionMaxTokensEnv overrides defaultVisionMaxTokens: an escape hatch for a
+// backend that thinks past even a generous default, or for capping a metered
+// one. The describe call is one hop deep inside a request, so it has to be
+// adjustable without a rebuild. Values below 1 are ignored.
+const visionMaxTokensEnv = "TINGLY_VISION_MAX_TOKENS"
```
0x0079, line 74: It is ok for current PR, minor and useful.
And we may let this configurable in future which may extend vision proxy rule flag config.

0x0079 (approved):
LGTM

PR #1640, opened 2026-08-25
fix(visionproxy): anchor the latest-turn test on the last image-bearing message

cbrown350:
## Summary

The vision proxy decides which images to describe by asking whether an image sits in the *last* message. Claude Code appends a `<system-reminder>` as its own system-role message after every tool result, so the image of the turn being answered is never last — it was replaced with the "omitted from history" marker before any model saw it, and the caller answered from nothing.

## Key Changes

- **The latest-turn test anchors on content, not position**: `lastIdx` becomes the index of the last message that can carry an image from the caller (user for the Anthropic shapes, user or tool for Chat Completions, user items for Responses) instead of `len(messages)-1`. Trailing messages that cannot carry an image no longer move the anchor, and when nothing qualifies the anchor stays at `len-1`, so behaviour is unchanged for every request whose final message is the turn.

## Notes

- Reproduced against Claude Code through a `claude_code` rule whose vision proxy is configured: a request answering a `Read` arrives as `user / system / assistant(tool_use) / user(tool_result+image) / system`, image at index 3 of 5.
- Measured end to end, asking for a registration painted on an aircraft in a photograph: **0/6 correct before, 4/4 after**. The failures were the dangerous kind — three replies invented a plausible registration rather than reporting a missing image, because the model was handed the marker text and answered around it.
- Genuine history stays elided: a test asserts an image from an earlier user turn still gets the cheap marker and costs no upstream call.

### DIFF
--- a/internal/vision/visionproxy/vision_proxy.go
+++ b/internal/vision/visionproxy/vision_proxy.go
@@ -45,6 +45,32 @@ const imageUnavailableText = "[image: (description unavailable)]"
 // message are sent through the vision upstream for description.
 const imageHistoricalText = "[image: (omitted from history)]"
 
+// latestImageAnchor reports the index the "latest turn" test should compare
+// against: the last message that can carry an image from the caller, rather
+// than the last message outright.
+//
+// Clients append trailing messages the caller never sees as part of the turn.
+// Claude Code emits a `<system-reminder>` as its own system-role message after
+// every tool result, so a request answering a tool call arrives as
+//
+//	user / system / assistant(tool_use) / user(tool_result+image) / system
+//
+// Anchoring on len-1 makes the image of the turn *in flight* test as history,
+// and it is replaced with imageHistoricalText before any model sees it — the
+// caller then answers about an image that was never described and never sent.
+//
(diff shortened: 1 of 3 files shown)


internal/vision/visionproxy/vision_proxy.go
```diff
@@ -45,6 +45,32 @@ const imageUnavailableText = "[image: (description unavailable)]"
 // message are sent through the vision upstream for description.
 const imageHistoricalText = "[image: (omitted from history)]"
 
+// latestImageAnchor reports the index the "latest turn" test should compare
```
FFengIll, line 48: I got this case, and thanks for the solution.

This only problem is `cache` - we do not cache all images, so each time the image change we break llm token cache in provider/vendor side.

Maybe we can improve this in future, but I leave this as a discussion here.

FFengIll (approved):
Thanks for the PR. Approved.

FFengIll:
In past version, tb do not support cache for vision proxy, and it makes trouble here.

https://github.com/tingly-dev/tingly-box/pull/1696 now add the cache may resolve both image missing and prompt cache   missing problems - it will ignore the position and process all images in messages.

PR #1647, opened 2026-08-26
bugfix(routing): degrade instead of failing rules when runtime state empties the candidate set

0x0079:
## Summary

A rule whose only active service failed repeatedly (429/auth marked it unhealthy) could be force-evicted from routing — every request then died with "no service available for rule" instead of surfacing the real upstream error. 

This held whenever the pipeline's candidate set emptied while the rule still had active services: an inactive sibling service row, or a smart-routing partition keeping HealthStage's union-level degrade guard from firing.

## Key Changes

- **Candidate seeding**: inactive services no longer enter the pipeline's candidate set, so they can't mask the all-unhealthy degrade guard and get the only active service eliminated.
- **Single-owner degrade invariant**: `ServiceSelector.Select` restores the rule's active base pool between stages whenever a narrowing comes back empty; per-stage compensations are removed, and an empty set at the terminal stage again means exactly "zero active services configured".
- **Affinity pin scope**: an empty candidate set now declines a pin instead of trusting it blindly, closing a leak where a partition-only service could serve a non-matching request.

## Testing

- Exhaustive invariant test: 9376 combinations of Active × health × breaker × partition × match × pin × tactic through the real stack; reverting the fixes fails 296+ of them.
- Targeted repro tests for each emptying path, plus a new `noeviction` harness scenario (`harness lb --example noeviction`) that reproduced the reported failure pre-fix.

---
_Generated by [Claude Code](https://claude.ai/code/session_017HDs4vzeTdRvtJ2P7Z5fpp)_

### DIFF
--- a/cli/harness/README.md
+++ b/cli/harness/README.md
@@ -281,7 +281,7 @@ upstreams. It runs the real `ServiceSelector.Select → dispatchWithPriorityFail
 deterministic breaker clock, so recovery is exercised without sleeping.
 
 ```bash
-./harness lb --example cascade        # cascade | flat | grid | single | regression | ratelimit | authflip | crossmodel | halfopen | degrade | inactive | withintier | multiaffinity
+./harness lb --example cascade        # cascade | flat | grid | single | regression | ratelimit | authflip | crossmodel | halfopen | degrade | inactive | noeviction | withintier | multiaffinity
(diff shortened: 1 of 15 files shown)


andreasfoo:
LGTM

PR #1696, opened 2026-09-02
feat(vision-proxy): cache image descriptions per session to cut cost and stabilize prompt caching

### LINKED ISSUE #1692: [TEP-1692] Add image <-> text cache in Vision Proxy
I got this case, and thanks for the solution.

This only problem is `cache` - we do not cache all images, so each time the image change we break llm token cache in provider/vendor side.

Maybe we can improve this in future, but I leave this as a discussion here.

_Originally posted by @FFengIll in https://github.com/tingly-dev/tingly-box/pull/1640#discussion_r3909943372_

FFengIll:
## Summary
Vision Proxy re-described every image in the latest message on every request, even byte-identical repeats within a retry/failover burst — wasting vision-model calls and, since descriptions are non-deterministic, breaking downstream prompt-prefix caching. 

Adds a session-scoped describe cache so repeats within the same conversation are free and consistent, and brings the design docs up to date with it.

Closes #1692.

## Key Changes

- **Describe cache**: images are cached by `(session, provider, model, image-hash)` — a resent image within a conversation (retry, failover, repeated tool call) skips the vision call and gets byte-identical replacement text. Content is xxhash64 + length of the base64 text (with media type) or of the URL; a cryptographic digest cost a 30-screenshot session a quarter second per turn.
- **Session-scoped, not global**: keyed per conversation so a bad description can't get stuck forever, and unrelated sessions never cross-contaminate each other's descriptions.
- **Historical images can reuse a real description**: if a historical image's content was already cached (e.g. it was last turn's latest message), it gets the real description instead of a marker.
- **Model-switch safety**: provider+model are part of the cache key, so reconfiguring the vision service invalidates stale entries automatically instead of serving another model's answer.
- **Stored in `tingly.db`, single tier**: descriptions live in a `vision_descriptions` table, so a restart does not re-describe every image a live conversation carries. No in-memory LRU in front of it — a SQLite point lookup is cheaper than hashing the image, and one source of truth beats promotion/write-through logic. No age expiry; only a 100k-row LRU ceiling as a safety valve. The session key is `source:value` (no client-IP backup) so a network change keeps hitting. A bounded in-process map is the fallback when no database is available.
- **Bounded, newest-first describe instead of a "latest message" test**: position no longer decides whether an image is eligible, only how misses are ranked. Every image checks the cache; misses are reversed to newest-first, repeated images fold into one describe spliced into every position, and the first `TINGLY_VISION_DESCRIBE_LIMIT` (default 8) are described and cached, in that order; older ones get a deferral marker this turn. A fully uncached history converges to fully described within a few turns while each request stays bounded. `latestImageAnchor` and the per-protocol `lastIdx` plumbing are removed.
- **Per-describe timeout**: each upstream call is bounded by `TINGLY_VISION_DESCRIBE_TIMEOUT` (default 60s) instead of running on the request context alone; a hung vision model no longer holds the request until the client gives up.
- **Negative cache for failures**: a failed (or timed-out) describe is remembered in memory for 10 minutes; within that the image is fail-stripped without a retry and without a describe slot, so a permanently failing image cannot hold a slot every turn and starve older images. Never persisted; a transient failure retries after the TTL and a recovery is cached.
- **No usable service**: every image is fail-stripped uniformly, without going through the bound, so none is told it is merely "not yet described".
- **Scenario matrix**: `vision_scenario_test.go` drives the proxy through `Service.Apply` with Claude Code's real multi-turn message shape and a vision client whose wording changes on every call, so any re-describe shows up as a text change — tool loop, restart, model switch, proxy enabled mid-conversation, upstream outage then recovery, duplicate screenshot, permanently bad image. `.design/vision-proxy.md` §11 is the matrix those tests implement, plus the scenarios judged to need nothing.
- **Docs**: `.design/vision-proxy.md` and the package README updated end-to-end to match current code and describe the cache design (key rationale, storage, retention, known limitations).

## Notes

- Rebased on #1640; its trailing-system-message fix is subsumed by the bounded rule (the turn in flight wins the slot by position), and its tests are kept in that form.
- The cache is per gateway instance (its own `tingly.db`), not shared across instances.
- Concurrent requests carrying the same new image are deliberately not coalesced, to keep description diversity.
- IP-fallback sessions (no session header, no `metadata.user_id`) share one cache scope per client IP, as `ResolveSessionID` already does for affinity; documented, not changed.
- URL images are identified by URL text: a per-request presigned URL is a new image every turn.


cbrown350:
Approved and agreed the cache helps — but to be precise, it only ignores image position on a **cache hit**. The miss path still anchors "latest" on `len(messages)-1`, and in the Claude Code flow (`user / system / assistant(tool_use) / user(tool_result+image) / system`) the image's *first* appearance is always a miss: it tests as historical, gets the "omitted from history" marker, and never earns a cache entry — so the image-missing bug persists on `feat/image_cache` as-is.

#1640 anchors the latest-turn test on the last image-bearing message so the first sight of the image is described (and then cached by your PR). They compose cleanly; suggest merging #1640 first and threading `latestImageAnchor` into your `collect*` `lastIdx` lines on rebase.

FFengIll:
> #1640 anchors the latest-turn test on the last image-bearing message so the first sight of the image is described (and then cached by your PR). They compose cleanly; suggest merging #1640 first and threading `latestImageAnchor` into your `collect*` `lastIdx` lines on rebase.

You are absolutely right.
We resolved two different issues.
Thanks for the revision.

0x0079:
Rebased onto `main` with #1640 in, so `collect*` now anchors on `latestImageAnchor` and the first sight of an image in the Claude Code flow is described and then cached — the composition you described.

On top of that the branch now carries a persistent tier: the memory LRU fronts a `vision_descriptions` table in `tingly.db`, so a gateway restart or an LRU eviction no longer re-describes (and re-words) every image a live conversation carries. No age expiry, just a 100k-row LRU ceiling as a safety valve. The PR description is updated to match.

---

PR #1768, opened 2026-09-18
fix(ui): make copy work in non-secure contexts (plain HTTP / Tailscale)

### LINKED ISSUE #1746: [BUG] Copy buttons fail ("Failed to copy to clipboard") when the Web UI is served over plain HTTP from a non-localhost host (Tailscale/LAN)
### Version

v0.260910.1

### OS & Architecture

Linux x86_64

### Category

UI

### Scenario

Tingly-Box is deployed on a remote server (started with `--host 0.0.0.0`) and I open the Web UI from another machine over Tailscale at `http://100.x.y.z:12580` (plain HTTP, non-localhost host). Any copy button — token, URL, or a config snippet — fails: it shows **"Failed to copy to clipboard"** (or silently does nothing). The same buttons work when the UI is opened at `http://localhost:12580` on the server itself.

**Expected:** copy buttons work regardless of how the UI is reached.
**Actual:** copy fails whenever the page is served in a non-secure browsing context (plain HTTP to a non-`localhost` host).

### Steps to Reproduce

1. Run on server A: `tingly-box start --host 0.0.0.0`.
2. From machine B (connected to A over Tailscale, or any LAN host), open `http://<A-tailscale-or-lan-ip>:12580`.
3. Click any copy button (e.g. the token on the Access Control page, or a config snippet on a scenario page).
4. A "Failed to copy to clipboard" error appears; nothing is copied.

### Logs / Error Output

```shell
# Browser console (non-secure context):
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'writeText')
# or navigator.clipboard === undefined
```

Root cause: `navigator.clipboard` is only available in a **secure context** (HTTPS, or `http://localhost` / `http://127.0.0.1`). When the UI is served over plain HTTP to a non-`localhost` host (Tailscale IP, LAN IP), `navigator.clipboard` is `undefined`, so every copy path that calls `navigator.clipboard.writeText(...)` fails. The copy helpers have no fallback:

- `frontend/src/hooks/useFunctionPanelData.ts:33-40` — `copyToClipboard` catches the throw and shows `"Failed to copy to clipboard"` (the message reported above).
- `frontend/src/hooks/useCopyFeedback.ts:11` — `navigator.clipboard.writeText(text).then(...)` with no `.catch` and no fallback (silent failure).
- Plus ~15 direct `navigator.clipboard.writeText` call sites across the frontend (e.g. `pages/system/AccessControl.tsx`, `pages/SharingKeysPage.tsx`, `components/ApiKeyTable.tsx`, `components/OAuthDialog.tsx`, `components/rule-card/utils.ts`, `components/ModelRequestHeader.tsx`, ...).

Suggested fix: a shared clipboard helper that tries `navigator.clipboard.writeText` when available and falls back to the legacy `document.execCommand('copy')` (via a temporary off-screen `<textarea>`) when it is not, so copy works in non-secure contexts; route the copy hooks and all direct call sites through it.

MingxiYe:
## Summary

Copy buttons in the Web UI fail when the page is served over plain HTTP to a **non-localhost** host (a LAN IP, or a Tailscale `100.x` address). `navigator.clipboard` is only defined in a secure context (HTTPS, or `http://localhost` / `127.0.0.1`), so every copy path throws:

- `hooks/useFunctionPanelData.ts` catches it and shows **"Failed to copy to clipboard"**
- `hooks/useCopyFeedback.ts` had no `.catch` and silently no-ops
- the ~15 direct `navigator.clipboard.writeText` call sites throw

## Fix

- Add `frontend/src/utils/clipboard.ts` `copyText()`: uses `navigator.clipboard.writeText` in a secure context, otherwise falls back to the legacy `document.execCommand('copy')` via an off-screen `<textarea>`; rejects if neither path can copy.
- Route the copy hooks (`useCopyFeedback`, `useFunctionPanelData`) and all direct `navigator.clipboard.writeText` call sites through it.
- Promotes the correct fallback that already existed locally in `components/rule-card/utils.ts` into the shared helper (removes the duplicate).

## Test plan

- `pnpm typecheck`: no new errors (40 before this change, 40 after — the repo's pre-existing legacy errors are unchanged).
- `pnpm lint`: no new findings (only the repo's pre-existing React-pattern warnings remain).
- Secure-context behavior (localhost / HTTPS) is unchanged — it still uses `navigator.clipboard`.
- The non-secure fallback path (plain HTTP to a non-localhost host, e.g. over Tailscale) runs `execCommand('copy')`. That branch cannot be exercised from a headless localhost/HTTPS context (where `navigator.clipboard` is present), so it is verified by review here rather than an automated run.

Fixes #1746


frontend/src/utils/clipboard.ts
```diff
@@ -0,0 +1,34 @@
+/**
```
FFengIll, line 1: hwo about to update navigator.clipboard directly ?  if we can not, we may create anothe rnew clipboard.

PR #1776, opened 2026-09-18
bugfix(protocol,codex): fix Codex prompt-cache invalidation and 400s

FFengIll:
## Summary
Requests forwarded to the Codex/ChatGPT backend re-serialized conversation history differently from turn to turn, invalidating Codex's own prompt cache and burning tokens on long sessions. 

Stabilizing that serialization shape then exposed a latent Responses-API bug unconditionally — assistant text was tagged as `input_text` instead of `output_text`, returning 400s on any multi-turn conversation routed through Codex. Both issues are now fixed end-to-end.

## Key Changes

- **Codex prompt-cache stability**: Converted request bodies retain the same shape as a client's cache breakpoint advances between turns, so Codex's upstream cache hits instead of missing from the first divergent byte.
- **Cache affinity hint**: Derives a stable `prompt_cache_key`/`session-id` from Claude Code's own session id and forwards it to Codex, closing a gap where converted traffic carried no affinity hint at all.
- **Assistant content-type fix**: Assistant text now serializes as `output_text`/`output_message`, not `input_text`, in both conversion directions (Anthropic/Chat → Responses and back) — fixes the 400s.
- **Vendor wire-shape selection**: Picks the OpenAI Chat wire shape (array vs. string content, explicit cache fields) per upstream vendor instead of assuming every OpenAI-compatible provider accepts the same shape.
- **Cross-request cache test suite**: New `cache_prefix` harness section validates that consecutive requests within a single conversation share a serialized prefix, not just that each request is individually valid.

## Notes

- Prompt-cache breakpoints have no `output_text` equivalent, so `cache_control` on assistant text is dropped when converting to Responses — an inherent API limitation, documented in `protocol-responses.md`.
- `cache_prefix` proves shape stability, not per-role wire-schema legality, so it would not catch a recurrence of the content-type bug — noted in `harness-matrix.md` §10.4.


internal/protocol/request/openai_chat_to_openai_responses.go
```diff
@@ -235,28 +235,39 @@ func convertChatUserMessageToResponses(userMsg *openai.ChatCompletionUserMessage
...
+//
+// Emits an output_message item (responseOutputMessageWithText), matching the
+// native shape the Anthropic->Responses path emits for assistant text (see
+// convertV1AssistantMessageToResponsesInput). The Responses API also accepts
+// assistant content as a plain-string message item, but a mismatched
+// content-part form on that path is what caused the input_text/output_text
+// bug this converter is the twin of (cache_control.go) — using the one
+// unambiguous shape here keeps the two conversion paths from producing two
+// different but both-valid wire forms for the same input.
 func convertChatAssistantMessageToResponses(assistantMsg *openai.ChatCompletionAssistantMessageParam) []responses.ResponseInputItemUnionParam {
```
FFengIll, line 256: agree to unify.

internal/protocol/request/cache_control.go
```diff
@@ -24,29 +93,18 @@ func applyFirstResponsesCacheBreakpoint(req *responses.ResponseNewParams) {
...
-			if content.OfString.Valid() && content.OfString.Value != "" {
-				text := &responses.ResponseInputTextParam{
-					Text:                  content.OfString.Value,
-					PromptCacheBreakpoint: responses.NewResponseInputTextPromptCacheBreakpointParam(),
-				}
-				content.OfString = param.Opt[string]{}
-				content.OfInputItemContentList = responses.ResponseInputMessageContentListParam{
-					{OfInputText: text},
-				}
+		if item.OfMessage == nil {
```
FFengIll, line 96: no more simple text message, all to part.

PR #1824, opened 2026-09-24
fix(oauth): send the login's account id as metadata account_uuid, random only as fallback

0x0079:
## Summary
OAuth logins minted a random `OAuthDetail.UserID` even when the token exchange had already captured the real account id, so `metadata.user_id.account_uuid` sent upstream was a made-up value instead of the account's real one.

## Key Changes

- **Login**: `oauthUserIDFromToken` uses `token.Metadata["account_id"]` when present, random uuid only as fallback — applies to every OAuth issuer.
- **Existing providers**: the read site (`transform.WithProvider`) prefers `OAuthDetail.ExtraFields["account_id"]` over `UserID`, so providers logged in before this change send the real account uuid on their next request with no stored-data migration.

## Notes
`UserID` also keys the in-memory OAuth `TokenStorage`; two providers sharing one account now share that slot. Nothing on the request path reads the store (only the deprecated `?user_id=` endpoints do), so this is harmless.

https://claude.ai/code/session_01BaqrQHd5R1M839cEUigdks


andreasfoo (approved):

andreasfoo:
LGTM, non-empty then go into context

